PR #179, opened 2021-02-09
Assert MACD isStable after age has passed all intervals

nullobject:
The MACD indicator should only be stable after all intervals have expired. It must wait until the age reaches the largest of the intervals.

This allows for any intervals, like (26, 12, 9) or (5, 2, 9), etc.

### DIFF
--- a/src/MACD/MACD.ts
+++ b/src/MACD/MACD.ts
@@ -30,7 +30,7 @@ export class MACD {
   }
 
   get isStable(): boolean {
-    return this.age >= this.config.longInterval;
+    return this.age >= Math.max(this.config.longInterval, this.config.shortInterval, this.config.signalInterval);
   }
 
   update(_price: BigSource): void {


bennycode:
But the `longInterval` is supposed to be the largest one. Why do you need to use `Math.max` to find out that `shortInterval` is shorter than `longInterval`?

nullobject:
@bennycode Fair enough, this was just to treat them all equally.

But `signal` can definitely be longer than `longInterval`.

bennycode:
I would argue that it makes sense to have a `signalInterval` larger than the `longInterval` (slow line). [Typical MACD settings](https://www.dailyfx.com/education/moving-average-convergence-divergence/macd-settings.html) don't work like this. However, it's possible and I see your point now. Thank you for the PR. I will merge it and release it with the next version. :)

nullobject:
@bennycode Thanks buddy.

`(long:26, short:12, signal:9)` are definitely the traditional values, but I have seen `(long:5, short:2, signal:9)` used in some cases.

PR #333, opened 2021-10-15
Types are optional.

hannesvdvreken:
If someone wants to use types to use trading-signals with big numbers, they need to add the types dependency themselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,11 +3,11 @@
     "url": "https://github.com/bennycode/trading-signals/issues"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


bennycode:
Thanks for your insights. I want to make it easy for people to use "trading-signals" and using it with Big.js is a [core functionality]( https://github.com/bennycode/trading-signals#usage). That's why I don't want to put the burden on third-party developers to install additional types themselves. They will get everything out of the box with "trading-signals". Types hardly take up any storage space and are therefore negligible during the download process. But thanks for bringing this up!

PR #348, opened 2021-11-07
feat(EMA): Add faster EMA implementation

patotoma:


bennycode:
Thanks @patotoma for your contribution. I am happy to merge it and I will add the `FasterEMA` to the [performance benchmark](https://github.com/bennycode/trading-signals/blob/main/src/start/startBenchmark.ts). 👍

PR #393, opened 2022-01-02
Draft: Native Implementation for NodeJS

Moumouls:
## Checklist for new indicators

- [ ] Indicator is documented and implemented by extending `BigIndicatorSeries`
- [ ] A "faster" version of this indicator is implemented by extending `NumberIndicatorSeries`
- [ ] Tests for `getResult` are present
- [ ] Tests for highest and lowest result caching are present (if not already existent for base class)
- [ ] Tests for error handling are present (if not already existent for base class)
- [ ] Indicators (standard & faster version) are exposed in `src/index.ts`
- [ ] Indicators (standard & faster version) are added to `startBenchmark.ts`
- [ ] Indicator is listed in `README.md`


Moumouls:
Hi @bennycode i'll push my work on this draft PR.

NAPI-RS suggest to push a new npm package for the `native` version to support binaries variations correctly.

Since the native version could have some little difference on interfaces and also it's not suitable for browser i can suggest to publish the `native` directly under a new npm package like `@trading-signals/native`.

My goal here with the `native` directory on this repo is to recycle all tests. A real mono repo could be more appropriate for the current use case but i think it's currently overkill.

Moumouls:
I tried the benchmark script i got an interesting result:

```
DEMA x 39.91 ops/sec ±1.13% (51 runs sampled)
NativeDEMA x 30,283 ops/sec ±0.88% (85 runs sampled)
FasterDEMA x 70,241 ops/sec ±2.48% (88 runs sampled)
```

I also tried a new console.time on MACD test and now

```ts
      const arr = Array.from({length: 100000}).map(() => Math.random() * 100);
      const arr2 = [...arr];
      console.time('fast');
      arr.forEach(v => {
        fasterMACD.update(v);
      });
      console.timeEnd('fast');
      console.time('native');
      arr2.forEach(v => {
        nativeMACD.fastUpdate(v);
      });
      console.timeEnd('native');
```

```
    fast: 11 ms
    native: 23 ms
```

Moumouls:
Okay @bennycode  i'l close this one, my first results on the discussion are wrong, i think i missed something like disabling nyc.

The benchmark script is right and you were right too, the communication between node and native bin is not fast enough.

PR #666, opened 2024-04-15
fix replace feature in TR

ozum:
The replace feature in the `update` of the `TR` indicator was not working correctly, because it does not handle `previousCandle` when replacing the result.

I added a new attribute to store two previous candles that replace the previous candle when `replace` is `true`. I also included a test to ensure that replace feature is working as expected.


bennycode:
Thank you @ozum for making this important contribution! 👍 I released your fix with trading-signals v5.0.1 ☀️

PR #676, opened 2024-05-06
Remove console logs in Indicator.ts

lums31:
Minor: Remove console logs, this floods the browsers console tab  😬


bennycode:
Very good catch, thank you!!

PR #677, opened 2024-05-07
Adding weighted moving average

AkshatGiri:
## Weighted Moving Average Checklist

- [x] Indicator is documented and implemented by extending `BigIndicatorSeries`
- [x] A "faster" version of this indicator is implemented by extending `NumberIndicatorSeries`
- [x] Tests for `getResult` are present
- [x] Tests for highest and lowest result caching are present (if not already existent for base class)
- [x] Tests for error handling are present (if not already existent for base class)
- [x] Indicators (standard & faster version) are exposed in `src/index.ts`
- [x] Indicators (standard & faster version) are added to `startBenchmark.ts`
- [x] Indicator is listed in `README.md`

Feature requested in discussion - https://github.com/bennycode/trading-signals/discussions/674

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -60,6 +60,7 @@ All indicators can be updated over time by streaming data (prices or candles) to
 1. Stochastic RSI (STOCHRSI)
 1. True Range (TR)
 1. Wilder's Smoothed Moving Average (WSMA / WMA / WWS / SMMA / MEMA)
+1. Weighted Moving Average (WMA)
 
 Utility Methods:
 
--- a/src/WMA/WMA.test.ts
+++ b/src/WMA/WMA.test.ts
@@ -1,0 +1,134 @@
+import {Big, FasterWMA, NotEnoughDataError, WMA} from '../index.js';
+import {describe} from 'vitest';
+
+describe('WMA', () => {
+  describe('prices', () => {
+    it('does not cache more prices than necessary to fill the interval', () => {
+      const wma = new WMA(3);
+      const fasterWMA = new FasterWMA(3);
+      wma.update(1);
+      fasterWMA.update(1);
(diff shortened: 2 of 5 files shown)


docs/README.md
```diff
@@ -60,6 +60,7 @@ All indicators can be updated over time by streaming data (prices or candles) to
 1. Stochastic RSI (STOCHRSI)
 1. True Range (TR)
 1. Wilder's Smoothed Moving Average (WSMA / WMA / WWS / SMMA / MEMA)
+1. Weighted Moving Average (WMA)
```
bennycode, line 63: Is this different to "Wilder's Smoothed Moving Average" (also shortened with "WMA")?

AkshatGiri: Yup it's different. 

Weighted Moving Average is more commonly abbreviated to WMA, compared to Wilders Smoothed Moving Average which typically goes by WSMA or SMMA in my experience. 

Didn't want to remove the WMA abbreviation from the WSMA tho since I searched it up and seems like it's still used sometimes for Wilders as well.

What do you think, should we keep it or remove it?

bennycode: Let's remove the WMA abbreviation from Wilder's Smoothed Moving Average then to avoid confusion. :)

bennycode:
Thank you very much for your contribution @AkshatGiri! 👏

I verified that the output from WMA & WSMA are different and I also fixed a little issue with the caching of the lowest & highest result: https://github.com/bennycode/trading-signals/pull/684/

You can use WMA now in trading-signals@5.0.4. 

Let me know how it goes! 💫

Best,
Benny

AkshatGiri:
Thanks for the quick merge @bennycode

PR #1062, opened 2026-04-29
feat(trading-strategies, trading-signals-docs): add TrailingStopStrategy

bennycode:
## Summary

- Adds `TrailingStopStrategy` — an exit-only trailing stop that extends `Strategy` directly. The strategy attaches to an existing base balance (manual position, another strategy, or post-restart) and never opens positions itself.
- Tracks the highest candle high seen since attach and emits a sell-all advice when `candle.close <= peak * (1 - trailDownPct/100)`. The advice re-emits on every candle until `onFill` brings the position to zero, then triggers `onFinish` for session teardown.
- Configuration:
  - `trailDownPct` — exit threshold below the running peak (defaults to `"10"`)
  - `trailUpPct` (optional) — hysteresis on peak ratcheting; the peak only updates when `candle.high >= peak * (1 + trailUpPct/100)`. Useful in choppy markets to avoid micro-ticks resetting the trail.
  - `pivotPrice` (optional) — explicit initial peak; overrides the default seed from the attach candle's `high`.
  - `exitOrder` — `"limit"` (default) places the sell at the trail target; `"market"` guarantees fill at the prevailing price.
- Registered in `StrategyRegistry` so `createStrategy()` and `StrategyMonitor` can instantiate it by name.
- State (`exited`, `positionSize`, `peakPrice`, `exitReason`, `exitLimitPrice`) round-trips through `restoreState` with a type-predicate guard that enforces cross-field invariants and falls back to `defaultState()` on a malformed blob. `peakPrice === '0'` doubles as the "not yet attached" sentinel.

## Docs site

- Surfaces `TrailingStopStrategy` in the interactive backtest page (`packages/trading-signals-docs/pages/backtest.tsx`) with default config `{trailDownPct: '10', exitOrder: 'limit'}`. The description nudges users to set `initialBase > 0` since the strategy is exit-only.

## README

- Adds glossary bullets to `packages/trading-strategies/README.md` defining **day trader**, **swing trader**, and **position trader** in the existing Domain Knowledge section.

## Notes

- The strategy is intentionally *not* anchored to cost basis — there is no break-even check or "wishful limit at break-even" fallback. A breach below the trail target always exits, regardless of P&L.
- If you attach mid-drawdown and don't supply `pivotPrice`, the peak seeds from the attach candle's `high`, which may be below the historical high. Use `pivotPrice` to set an explicit anchor when that matters.
- `TradingSession` only forwards fills for orders it placed, so external fills do not reach `onFill`. Position is set once at attach and only decreases via the strategy's own SELL fills.

### DIFF
--- a/packages/trading-signals-docs/pages/backtest.tsx
+++ b/packages/trading-signals-docs/pages/backtest.tsx
@@ -11,11 +11,13 @@ import {
   ScalpStrategy,
   MeanReversionStrategy,
   ProtectedStrategy,
+  TrailingStopStrategy,
   type BacktestResult,
   BuyOnceConfig,
   BuyBelowSellAboveConfig,
   type MultiIndicatorConfluenceConfig,
   type ScalpConfig,
+  type TrailingStopConfig,
 } from 'trading-strategies';
 import {DatasetSelector} from '../components/DatasetSelector';
 import {StrategyConfigurator} from '../components/StrategyConfigurator';
@@ -43,6 +45,8 @@ function createStrategy(strategyId: StrategyId, config: Record<string, unknown>)
       return new MeanReversionStrategy({config});
     case 'protection-only':
       return new ProtectedStrategy({config});
+    case 'trailing-stop':
+      return new TrailingStopStrategy(config as TrailingStopConfig);
   }
 }
 
--- a/packages/trading-signals-docs/utils/strategySchemas.ts
+++ b/packages/trading-signals-docs/utils/strategySchemas.ts
@@ -1,10 +1,10 @@
 import {z} from 'zod';
 import type {ExchangeCandle} from '@typedtrader/exchange';
-import {BuyOnceSchema, BuyBelowSellAboveSchema, CoinFlipSchema, MultiIndicatorConfluenceSchema, ScalpSchema, MeanReversionSchema, ProtectedStrategySchema, suggestScalpOffset} from 'trading-strategies';
+import {BuyOnceSchema, BuyBelowSellAboveSchema, CoinFlipSchema, MultiIndicatorConfluenceSchema, ScalpSchema, MeanReversionSchema, ProtectedStrategySchema, TrailingStopSchema, suggestScalpOffset} from 'trading-strategies';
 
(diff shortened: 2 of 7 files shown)


bennycode:
Thanks Copilot — addressing your four comments:

**1. Type guard cast pattern** — Keeping the `in`-narrowing form. `tsc --noEmit` is clean, and the project preference is to avoid `as Type` assertions where a type predicate or `in` narrowing works.

**2. `restoreState` order of operations** — Acknowledging the concern, but leaving as-is for this PR. The proxied state (used at runtime) is unaffected because it's bound to the constructor's state object, so a malformed persisted blob doesn't poison strategy behavior. Worth a follow-up to make `Strategy.#_state` symmetric with `ProtectedStrategy.restoreState`'s pattern, but that's a base-class concern beyond this strategy.

**3. External fills not reaching `onFill`** — Fixed. Removed the BUY branch in `onFill` entirely; the strategy is exit-only and only its own SELL fills are forwarded by `TradingSession`. Docs updated to match (commit 45bde1e).

**4. Cross-field invariants in the type guard** — Fixed. `isTrailingStopState` now rejects (a) `exited=true` with non-zero `positionSize` and (b) `peakPrice !== '0'` with `positionSize <= 0` and `exited=false` (the silent no-op state). Two new tests cover both cases (commit 45bde1e).

bennycode:
Second pass — addressing the new round of Copilot comments:

**5. Register in `StrategyRegistry`** — Fixed (commit 78a1e75). Added `TrailingStopStrategy` to the registry mapping so `createStrategy()` and `StrategyMonitor` can instantiate it by name. Skipped the registry test since none exists for the other strategies either.

**6. `restoreState` validate-first** — Fixed (commit 78a1e75). Refactored to validate the persisted blob first and forward either the validated state or `defaultState()` to `super.restoreState`, mirroring `ProtectedStrategy.restoreState`. A malformed blob can no longer reach `Strategy.#_state` even transiently.

**7. README glossary additions not in PR summary** — Fixed. Updated the PR description to call out the day/swing/position trader bullets added to the Domain Knowledge section.

PR #1063, opened 2026-04-29
feat(exchange, trading-strategies, messaging): Add strategy onMessage event

bennycode:
## Summary

- Adds a minimal pipe for strategies to surface user-facing messages without knowing anything about the messaging layer.
- `Strategy` base and `TradingSessionStrategy` interface gain an optional `onMessage(text: string)` callback. The session sets it in its constructor to relay any text the strategy emits as a session-level `'message'` event.
- `StrategyMonitor` subscribes to `'message'` and forwards the text to the user via the same platform layer used by fill / finish notifications. The user-facing format is `"<strategy> (<pair>): <text>"`.
- `TrailingStopStrategy` emits one message on the first trail breach. The per-candle re-emission of the SELL advice does not repeat the message — dedup uses the `exitReason` transition (`null` → set).

## Contract

```ts
// Strategy / TradingSessionStrategy
onMessage?: (text: string) => void;

// TradingSessionEventMap
message: [string];

// Session wires it
strategy.onMessage = text => this.emit('message', text);

// StrategyMonitor forwards every message to the user
session.on('message', text => this.#sendStrategyMessage(row, text));
```

Plain string. No severity, no category, no DB-level filter. Strategies are responsible for not being chatty — every emitted message reaches the user.

### DIFF
--- a/packages/exchange/src/trader/TradingSession.test.ts
+++ b/packages/exchange/src/trader/TradingSession.test.ts
@@ -470,6 +470,15 @@ it('cancels existing order before placing a new one', async () => {
       expect(onMessage).toHaveBeenCalledTimes(1);
       expect(onMessage).toHaveBeenCalledWith('hello from strategy');
     });
+
+    it('clears strategy.onMessage when the session stops', async () => {
+      await session.start();
+      expect(strategy.onMessage).toBeTypeOf('function');
+
+      await session.stop();
+
+      expect(strategy.onMessage).toBeUndefined();
+    });
   });
 
   describe('stop', () => {
--- a/packages/exchange/src/trader/TradingSession.test.ts
+++ b/packages/exchange/src/trader/TradingSession.test.ts
@@ -460,6 +460,18 @@ it('cancels existing order before placing a new one', async () => {
     });
   });
 
+  describe('strategy messages', () => {
+    it('relays strategy onMessage calls as session message events', () => {
+      const onMessage = vi.fn();
+      session.on('message', onMessage);
+
+      strategy.onMessage?.('hello from strategy');
+
+      expect(onMessage).toHaveBeenCalledTimes(1);
+      expect(onMessage).toHaveBeenCalledWith('hello from strategy');
+    });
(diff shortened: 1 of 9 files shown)


bennycode:
Addressing Copilot's four comments:

**1. `exitReason` overwrite mismatch** — Fixed (commit 8517593). `state.exitReason` is now only set on the first breach candle (mirrors the `onMessage` dedup we already had). The emitted advice's `reason` still reflects the latest candle.

**2. `onMessage` not cleared on session stop** — Fixed (commit 8517593). `TradingSession.stop()` now sets `this.#strategy.onMessage = undefined` so a stopped session can't keep emitting and the strategy doesn't retain a reference to the session. Added a TradingSession test covering the cleanup.

**3. Missing `StrategyMonitor` test for message routing** — Fixed (commit 8517593). Extracted the user-facing format to a `formatStrategyMessage` helper exported from `StrategyMonitor.ts`, and added `StrategyMonitor.test.ts` covering the format string and platform-prefix routing in the same minimal style as `WatchMonitor.test.ts`.

**4. Backward-compat / migration for `stopPrice`** — Resolved as won't-do. Per @bennycode there is no live persisted state to migrate, so adding migration plumbing would just be code that never runs. The strict predicate stays.

bennycode:
Round three on the new Copilot comments:

**5. `stopPrice` migration (re-raised, line 265)** — Resolved as won't-do, same rationale as before. There is no live persisted state predating this PR, so a migration path would be code that never runs. Predicate stays strict.

**6. Duplication in `#sendStrategyMessage` / fill / finish handlers** — Fixed (commit b7e7417). Extracted `#sendToAccount(row, message, label, level)` that owns the account lookup, prefix-based platform resolution, and warn-on-missing handling. Each caller now just builds its message and the helper handles routing + the success log.

**7. `logger.info` for every strategy message is too noisy** — Fixed in the same commit. The new helper takes a `level` argument; strategy messages log at `debug`, fill/finish stay at `info`.

PR #1071, opened 2026-05-05
fix(exchange): exit on market-data WebSocket close so orchestrator restarts

bennycode:
## Summary

A market-data `WebSocket` that drops after authentication used to leave the singleton holding a dead stream. Subscribers stopped receiving candles silently. A trailing-stop strategy on `INTC,USD` lost ~9 hours of intraday candles on May 5 because of this — `peakPrice` never ratcheted past May 1's high.

After exploring an in-place reconnect (kept singleton, re-attached handlers, replayed subscribes), code review surfaced a steady stream of subtle races: in-flight retries vs. teardown, cached dead connection during reconnect, post-auth error suppression hiding 405/409s on resubscribe, etc. Each fix exposed another.

This PR takes the simpler path: **fail loud, let Dokku restart**. After the stream authenticates, listen for transport `close` and `process.exit(1)`. The orchestrator restarts the worker, the full startup path runs again (re-establish stream, re-subscribe symbols, re-attach handlers), strategies restore from persisted state.

## Implementation

A single `stream.once('close', ...)` listener installed inside the existing `'authenticated'` handler. No new state, no new types, no new APIs.

```ts
stream.once('close', () => {
  console.error(`Market-data WebSocket closed for "${connectionId}". Exiting so the orchestrator restarts the worker.`);
  process.exit(1);
});
```

Diff: +28 / -154 vs. the previous reconnect-in-place attempt.

## Operational note

Cost is ~5 seconds of downtime per disconnect. Acceptable until drops become frequent enough to justify a real reconnecting transport — at which point the right path is to push the reconnect into `AlpacaStream` itself (the layer that wraps the native `WebSocket`), not back into the singleton above it.

## Verification path

After this lands and the worker redeploys: a transport drop will produce a `Market-data WebSocket closed for "<id>"` log line followed by a Dokku restart, then the usual `WebSocket streaming is authenticated` log line for a fresh connection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 *.bak
+# https://nextjs.org/docs/pages/guides/environment-variables#test-environment-variables
+.env*.local
 *.env
 *.tsbuildinfo
 *hot-update.js*
--- a/packages/exchange/src/exchange/alpaca/AlpacaWebSocket.ts
+++ b/packages/exchange/src/exchange/alpaca/AlpacaWebSocket.ts
@@ -9,6 +9,9 @@ const hasErrorCode = (error: unknown): error is {code: string | number} => {
 
 export interface AlpacaConnection {
   connectionId: string;
+}
+
+interface InternalConnection extends AlpacaConnection {
   stream: AlpacaStream;
 }
 
@@ -28,7 +31,7 @@ interface BarHandler {
  * connectionId across reconnects.
  */
 class AlpacaWebSocket {
-  #connections: Map<string, AlpacaConnection> = new Map();
+  #connections: Map<string, InternalConnection> = new Map();
   #symbols: Map<string, Set<string>> = new Map();
   #barHandlers: Map<string, BarHandler[]> = new Map();
   #connectionMeta: Map<string, {credentials: AlpacaStreamCredentials; source: string}> = new Map();
@@ -67,7 +70,7 @@ class AlpacaWebSocket {
     credentials: AlpacaStreamCredentials,
     source: string,
     reuseConnectionId?: string
-  ): Promise<AlpacaConnection> {
+  ): Promise<InternalConnection> {
     const singletonKey = `${credentials.apiKey}:${source}`;
 
     // Initial connect: reuse the existing connection if one already exists for these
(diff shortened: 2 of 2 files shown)


bennycode:
Addressing all five Copilot comments (commit a0a9fdd):

**1. Pre-auth `close` hang** — Fixed. Added a pre-auth `close` listener that rejects with a clear message. The `retry()` wrapper now sees the rejection and runs another attempt instead of hanging forever.

**2. Resubscribe error silenced** — Acknowledged in the log. The post-auth error handler is intentionally non-fatal (a 405/409 from a resubscribe is a server message, not a transport drop), but the reconnect success line now says `"resent N handler(s) and M subscription(s) — awaiting confirmation"` instead of implying bars are flowing. The actual confirmation comes via the separate `WebSocket streaming is subscribed` log line emitted by the stream's `subscription` handler.

**3. Reconnect-after-teardown** — Fixed. Added a `disconnect(connectionId)` method that evicts every map entry first, then closes the stream. The post-auth close listener now checks `#connectionMeta.has(connectionId)` before reconnecting, so the close event from a deliberate teardown is a no-op.

**4. Stale entry on permanent failure** — Fixed. Reconnect's catch block now calls `#evict(connectionId)`, which clears `#connections`, `#connectionMeta`, `#credentialToConnectionId`, `#symbols`, and `#barHandlers`. A subsequent `connect()` for the same credentials does a fresh handshake.

**5. Stale `.stream` on returned `AlpacaConnection`** — Fixed by narrowing the public type to `{connectionId}`. Verified no external caller reads `.stream` (`AlpacaExchange` only uses `connection.connectionId`); internal bookkeeping uses a private `InternalConnection extends AlpacaConnection { stream }`. Callers can no longer hold a stale stream reference because the type doesn't expose it.

Tests: exchange package 85/85 passing; tsc clean.

bennycode:
Pivoted on architecture (commit a0aaebc).

The reconnect-in-place approach kept exposing new races each round of review, so I replaced the entire singleton-with-reconnect machinery with a single post-auth `stream.once('close', () => process.exit(1))`. Dokku restarts the worker, the existing startup path re-establishes everything from scratch.

This makes all four of the latest comments moot — there's no in-flight reconnect to interrupt, no cached-during-reconnect window, no post-reconnect 405/409 silent-failure, and no half-implemented `disconnect()` that diverges from `AlpacaExchange.disconnect()`. The only state in play is what was there before the PR.

If drops become frequent enough that ~5s of restart per drop is operationally painful, the right next step is a reconnecting transport inside `AlpacaStream` (so the wrapped `WebSocket` is replaceable but the stream's identity is stable). That's a separate concern from making the singleton self-heal — and now the singleton stays the simple shape it had on `main`.

Diff: **+28 / −154** vs. the prior commit.

PR #1076, opened 2026-05-12
feat(exchange, messaging): observability for stuck market-data streams

bennycode:
## Why

Twice now the INTC trailing-stop strategy went silent because the market-data WebSocket dropped and the close event only surfaced in \`dokku logs\`. No signal made it out — no Telegram message, no error elsewhere. By the time anyone noticed, the strategy had been deaf for hours.

This PR closes the gap with two reactive signals — no time-based audit loop.

### 1. Forward close metadata so we can tell *why* the WS dropped

The native WebSocket \`error\` event is information-free by spec — it carries no error code, no reason. The actual cause of a drop only lives on the subsequent \`close\` event. \`AlpacaStream\` was throwing those fields away.

Now \`AlpacaStream\` emits \`{code, reason, wasClean}\` alongside \`close\`, and \`AlpacaWebSocket\` logs:

\`\`\`
Market-data WebSocket closed for "fb18a402-…" (code=1006, reason="", wasClean=false). Exiting…
\`\`\`

Standard close codes (1006 abnormal closure, 1011 server error, 1012 service restart, 4xxx Alpaca app-level) now distinguish a network blip from an Alpaca-side restart — the May 8 incident's log gave us nothing useful here.

### 2. Telegram alert before \`process.exit(1)\`

\`AlpacaWebSocket\` exposes an optional \`onClose\` handler invoked just before the process exit, with a **2-second deadline** so a hung Telegram call cannot keep the worker alive. \`startServer.ts\` wires it to a new \`MessagingPlatform.alertOperators(text)\` method that broadcasts to every configured owner; \`TelegramPlatform\` implements it with per-recipient failure isolation.

Operators learn about the drop in seconds via the same Telegram bot they already use, with the close code and reason in the message body.

## Files

- \`packages/exchange/src/broker/alpaca/api/AlpacaStream.ts\` — new \`AlpacaStreamCloseInfo\` type, forward CloseEvent fields.
- \`packages/exchange/src/broker/alpaca/AlpacaWebSocket.ts\` — log close code/reason/wasClean, optional \`onClose\` hook with 2s deadline.
- \`packages/exchange/src/broker/alpaca/index.ts\` — export the singleton so messaging can wire \`onClose\`.
- \`packages/messaging/src/platform/MessagingPlatform.ts\` + \`TelegramPlatform.ts\` — new \`alertOperators(text)\` broadcast method.
- \`packages/messaging/src/startServer.ts\` — wire \`alpacaWebSocket.onClose\` to \`alertOperators\`.

## Tests

\`exchange\` 82/82 passing, \`messaging\` 146/146 passing, both packages typecheck clean. Tooling-only test changes: two mock platforms gained the new \`alertOperators\` stub.

### DIFF
--- a/packages/messaging/src/platform/TelegramPlatform.ts
+++ b/packages/messaging/src/platform/TelegramPlatform.ts
@@ -784,6 +784,28 @@ export class TelegramPlatform implements MessagingPlatform {
     }
   }
 
+  /**
+   * Broadcast an operator alert to every configured owner. Per-recipient failures
+   * are logged but never thrown so a single bad chat ID can't block the rest of
+   * the fan-out. When no owners are configured (a dev-mode setup with the bot
+   * open to anyone), this is intentionally a no-op — there's nobody specific to
+   * notify.
+   */
+  async alertOperators(text: string): Promise<void> {
+    if (this.#ownerIds.length === 0) {
+      logger.warn('alertOperators called but no Telegram owner IDs are configured; dropping alert');
+      return;
+    }
+    const results = await Promise.allSettled(
+      this.#ownerIds.map(ownerId => this.sendMessage(`${PLATFORM_PREFIX}${ownerId}`, text))
+    );
+    for (const [index, result] of results.entries()) {
+      if (result.status === 'rejected') {
+        logger.warn({err: result.reason, ownerId: this.#ownerIds[index]}, 'Failed to deliver operator alert');
+      }
+    }
+  }
+
   get commandList(): string[] {
     return Array.from(this.#commands.keys()).map(cmd => `/${cmd}`);
   }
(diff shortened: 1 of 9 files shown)


packages/messaging/src/platform/TelegramPlatform.ts
```diff
@@ -784,6 +784,28 @@ export class TelegramPlatform implements MessagingPlatform {
...
+  async alertOperators(text: string): Promise<void> {
+    if (this.#ownerIds.length === 0) {
+      logger.warn('alertOperators called but no Telegram owner IDs are configured; dropping alert');
+      return;
+    }
+    const results = await Promise.allSettled(
+      this.#ownerIds.map(ownerId => this.sendMessage(`${PLATFORM_PREFIX}${ownerId}`, text))
+    );
+    for (const [index, result] of results.entries()) {
+      if (result.status === 'rejected') {
```
bennycode: @claude create a constant here

PR #1081, opened 2026-05-14
feat(messaging): add /accountEdit command to update API credentials

bennycode:
## Summary

Adds an `/accountEdit` command so an account's API key and secret can be rotated straight from the messenger — no need to `/accountRemove` and `/accountAdd`, which would change the account ID and break every linked watch, strategy and report.

> **Depends on #1082** — the wizard cleanup below is only safe once the Alpaca retry config is bounded.

### Plain command
`/accountEdit <id> <apiKey> <apiSecret>`

- Looks up the account via `getAccountOrError` (ownership-checked, same as `/accountRemove`).
- Keeps the account ID, name, exchange and paper mode — only the credentials change.
- Validates the new credentials against the exchange (`client.getTime()`) before persisting, mirroring `/accountAdd`.

### Telegram wizard
On Telegram, `/accountEdit` runs a conversations-based wizard mirroring the `/accountAdd` wizard:

- Pick an account from an inline keyboard.
- Send the new API key and secret as separate messages.
- **Both secret-bearing messages are deleted from the chat after receipt.**
- Credentials are validated against the exchange before being persisted.

### Restart of running sessions
After a successful edit, every running watch and strategy bound to that account still holds a broker client built from the *old* credentials. Both the command and the wizard now silently restart them via `restartAccountSessions`:

- `StrategyMonitor.restartForAccount` — persists each strategy's state, stops the session **without** cancelling open orders, then re-subscribes from the freshly loaded row (restores state, builds a client from the new credentials).
- `WatchMonitor.restartForAccount` — tears down and recreates each subscription; watches hold no accumulated state.

### Wizard cleanup
With the Alpaca retry config bounded in #1082, `client.getTime()` is guaranteed to settle on its own. The hand-rolled 10s `Promise.race` + `setTimeout` timeout — which also leaked its timer on the happy path — is removed from both the `accountAdd` and `accountEdit` wizards.

## Changes

- New `command/account/accountEdit.ts` (plain command) and `platform/wizards/accountEditWizard.ts` (Telegram wizard).
- New `Account.update(id, data)` model method (sets `updatedAt`).
- New `StrategyMonitor.restartForAccount` / `WatchMonitor.restartForAccount` + shared `restartAccountSessions` helper.
- Removed the `Promise.race` validation timeout from `accountAddWizard` and `accountEditWizard`.
- Registered the command/wizard in `startServer.ts`, `command/index.ts`, `TelegramPlatform.ts` and `wizards/shared.ts`.
- Added the command to the README table.

### DIFF
--- a/packages/messaging/src/command/account/accountEdit.ts
+++ b/packages/messaging/src/command/account/accountEdit.ts
@@ -1,0 +1,40 @@
+import {getBrokerClient} from '@typedtrader/exchange';
+import {Account} from '../../database/models/Account.js';
+import {assertId} from '../../validation/assertId.js';
+import {getAccountOrError} from '../../validation/getAccountOrError.js';
+
+// Request Example: "1 API_KEY API_SECRET"
+// Format: "<id> <apiKey> <apiSecret>"
+export async function accountEdit(request: string, userId: string) {
+  const parts = request.trim().split(' ');
+
+  if (parts.length !== 3) {
+    return 'Invalid format. Usage: /accountEdit <id> <apiKey> <apiSecret>';
+  }
+
+  const [idStr, apiKey, apiSecret] = parts;
+
+  try {
+    const accountId = assertId(idStr);
+    // Keeps the account ID, name, exchange and isPaper mode (and therefore all linked
+    // watches, strategies and reports) — only the credentials are replaced.
+    const account = getAccountOrError(userId, accountId);
+
+    const client = getBrokerClient({
+      exchangeId: account.exchange,
+      apiKey,
+      apiSecret,
+      isPaper: account.isPaper,
+    });
+    await client.getTime();
+
+    Account.update(accountId, {apiKey, apiSecret});
+
+    return `Account "${account.name}" (ID: ${accountId}) updated successfully. Connection test passed.`;
+  } catch (error) {
+    if (error instanceof Error) {
+      return `Error updating account: ${error.message}`;
+    }
+    return 'Error updating account';
+  }
+}
--- a/packages/messaging/src/command/account/accountEdit.ts
+++ b/packages/messaging/src/command/account/accountEdit.ts
@@ -3,13 +3,19 @@ import {Account} from '../../database/models/Account.js';
 import {assertId} from '../../validation/assertId.js';
 import {getAccountOrError} from '../../validation/getAccountOrError.js';
 
+interface AccountEditResult {
+  message: string;
+  /** Set only on success — signals the caller to restart the account's running sessions. */
+  accountId?: number;
+}
+
 // Request Example: "1 API_KEY API_SECRET"
 // Format: "<id> <apiKey> <apiSecret>"
-export async function accountEdit(request: string, userId: string) {
+export async function accountEdit(request: string, userId: string): Promise<AccountEditResult> {
   const parts = request.trim().split(' ');
 
   if (parts.length !== 3) {
-    return 'Invalid format. Usage: /accountEdit <id> <apiKey> <apiSecret>';
+    return {message: 'Invalid format. Usage: /accountEdit <id> <apiKey> <apiSecret>'};
   }
 
   const [idStr, apiKey, apiSecret] = parts;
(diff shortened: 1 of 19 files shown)


packages/messaging/src/command/account/accountEdit.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  if (parts.length !== 3) {
+    return {message: 'Invalid format. Usage: /accountEdit <id> <apiKey> <apiSecret>'};
+  }
+
+  const [idStr, apiKey, apiSecret] = parts;
+
+  try {
+    const accountId = assertId(idStr);
+    // Keeps the account ID, name, exchange and isPaper mode (and therefore all linked
```
bennycode: @claude remove this comment

bennycode: @claude which github app permissions do you need?

packages/messaging/src/platform/wizards/accountAddWizard.ts
```diff
@@ -64,6 +64,8 @@ export function makeAccountAddWizard() {
     const apiSecretChatId = apiSecretCtx.msg.chat.id;
     const apiSecretMessageId = apiSecretCtx.msg.message_id;
     if (apiSecret.startsWith('/')) {
+      // Also delete the already-received API key message — it carries a real credential.
```
bennycode: @claude remove the comment here pls

packages/messaging/src/command/account/accountEdit.ts
bennycode: @claude i gave you more permission, please remove the comment

bennycode (commented):
@claude rework the comments

packages/messaging/src/platform/wizards/accountAddWizard.ts
bennycode: @claude you are not doing anything

packages/messaging/src/service/StrategyMonitor.ts
```diff
@@ -158,6 +153,44 @@ export class StrategyMonitor {
...
+          await this.subscribeToStrategy(freshRow);
+        }
+        logger.info({strategyId: row.id, accountId}, 'Restarted strategy after account update');
+      } catch (error) {
+        logger.error({err: error, strategyId: row.id, accountId}, 'Failed to restart strategy after account update');
+      }
+    }
+  }
+
+  #persistStrategyState(strategyId: number, strategy: TradingStrategy): void {
```
bennycode: @claude name it "persistStrategyState"

packages/messaging/src/platform/wizards/accountEditWizard.ts
```diff
@@ -0,0 +1,110 @@
...
+        // and secret in its request.headers / config.headers fields.
+        const message = error instanceof Error ? error.message : 'Unknown error';
+        logger.error({message}, 'accountEdit wizard failed');
+        return {ok: false as const, error: message};
+      }
+    });
+
+    if (result.ok) {
+      await ctx.reply(`Account "${account.name}" (ID: ${account.id}) updated successfully. Connection test passed.`);
+      // Restart the account's running watches and strategies so they reconnect with the new
```
bennycode: @claude remove the comment

packages/messaging/src/platform/wizards/accountEditWizard.ts
```diff
@@ -0,0 +1,110 @@
...
+    await ctx.reply('Validating credentials…');
+
+    const result = await conversation.external(async () => {
+      try {
+        const client = getBrokerClient({exchangeId: account.exchange, apiKey, apiSecret, isPaper: account.isPaper});
+        await client.getTime();
+        Account.update(account.id, {apiKey, apiSecret});
+        return {ok: true as const};
+      } catch (error) {
+        // Log ONLY the message — the raw axios error carries the API key
```
bennycode: @claude not necessary to do this anymore as axios errors turn into a `SimplifiedHTTPError` without key leakage

packages/messaging/src/platform/wizards/accountAddWizard.ts
```diff
@@ -90,15 +92,7 @@ export function makeAccountAddWizard() {
...
-        // 10s timeout guards against the underlying Alpaca HTTP client's
-        // Infinity retry loop on network errors / 429s — without it a flaky
-        // validation could hang the wizard forever.
-        await Promise.race([
-          client.getTime(),
-          new Promise<never>((_, reject) =>
-            setTimeout(() => reject(new Error('Validation timed out after 10s')), 10_000)
-          ),
-        ]);
+        await client.getTime();
```
bennycode: @claude can we build a new function that bundles `getBrokerClient` and a `getTime` call? We may use this combination in other message commands already and this would save us from code duplication. We do the `getTime` call to have a guaranteed authorized client. How about naming this new function `getAuthenticatedBrokerClient`?

packages/messaging/src/command/account/accountEdit.ts
```diff
@@ -0,0 +1,49 @@
...
+    // watches, strategies and reports) — only the credentials are replaced.
```
bennycode: @claude remove the comment

PR #1085, opened 2026-05-15
feat(messaging): notify owner when strategies and watches start

bennycode:
## Summary
- On application boot, `StrategyMonitor.start()` and `WatchMonitor.start()` now send a short message to the owning account after each persisted strategy/watch resumes (`Strategy <id> started.` / `Watch <id> started.`).
- Wizards and `restartForAccount` paths are intentionally untouched — those flows already give the user direct feedback, so notifications fire only on boot resumes.


packages/messaging/src/service/StrategyMonitor.ts
```diff
@@ -225,6 +226,10 @@ export class StrategyMonitor {
     await this.#sendToAccount(row, message);
   }
 
+  async #sendStartNotification(row: StrategyAttributes): Promise<void> {
+    await this.#sendToAccount(row, `Strategy ${row.id} started.`);
```
bennycode: @claude wrap `${row.id}` in quotes.

PR #1095, opened 2026-05-28
test(trading-signals-docs): add Playwright e2e smoke tests for the homepage

bennycode:
## Summary

- Scaffolds Playwright under `packages/trading-signals-docs/e2e/` so we have a smoke check guaranteeing the docs site actually renders end-to-end
- One spec covers the hero, external NPM/GitHub links, and category navigation to `/indicators/trend/`
- Page objects use the `build<Name>Page` arrow-function factory convention — specs stay arrange/act/assert in ~3 lines
- `playwright.config.ts` boots `next dev` via `webServer`, with `APP_URL` env override for hitting deployed previews, JUnit reporter + screenshots/traces on CI, HTML reporter locally
- Vitest excludes `e2e/**` so the two suites don't overlap
- New `.claude/rules/docs-e2e.md` codifies when to reach for E2E vs unit and the page-object convention

## How to run

```sh
cd packages/trading-signals-docs
npx playwright install chromium   # first time only
npm run test:e2e
```

`test:e2e` is intentionally **not** wired into the default `npm test` flow — the browser download would slow down the standard test loop.


.github/workflows/run-docs-e2e.yml
```diff
@@ -0,0 +1,52 @@
+name: 'Run Docs E2E'
```
bennycode: @claude name it `Run E2E Docs` and rename the file to `run-docs-e2e.yml`

bennycode: @claude name it `Run E2E Docs` and rename the file to `run-e2e-docs.yml`

bennycode: @claude which permissions do you need? i can update your workflow file.

PR #1096, opened 2026-05-29
fix(trading-strategies): reconcile restored TrailingStop state against broker

bennycode:
## Summary

A live `TrailingStopStrategy` on this account (strategy id 8, attached to NVDA) has been emitting one Alpaca `422 qty must be > 0` error per minute since a worker restart on 2026-05-28. Root cause: on 2026-05-26 the trail's limit sell filled and closed the position on Alpaca, but the strategy's persisted state never updated to `exited: true`. After the May 28 restart, `TrailingStopStrategy.restoreState` reloaded `positionSize: "1"` and `exited: false`, and the strategy has been re-emitting a sell-all every candle against a phantom 1-share position. `TradingSession` resolves `AllAvailableAmount` against the live (now-zero) balance, ships qty=0 to Alpaca, and Alpaca rejects.

The most plausible cause of the missed persistence is an unclean shutdown between the fill arriving and `onFill` writing state — supported by the `connection limit exceeded` errors that appear in the May 28 startup logs (Alpaca still counted the previous worker's WebSockets as live).

This PR adds a one-shot reconciliation check on the first candle after restore:

- If the broker reports `baseBalance < persistedPositionSize`:
  - `baseBalance == 0` → mark strategy `exited: true`, set `positionSize = '0'`, emit an `onMessage` explaining what happened, and call `onFinish` so the runtime tears the session down.
  - `0 < baseBalance < positionSize` → partial drift; shrink `positionSize` to the live balance and continue trailing.
- Otherwise: no change.

The `#reconciled` flag is intentionally **not persisted** — every fresh process should reconcile once, because the persisted state could have been written before a fill arrived.

## Test coverage

`packages/trading-strategies/src/strategy-trailing-stop/TrailingStopStrategy.test.ts`:
- `reconciles to exited when broker reports zero base balance after restore`
- `shrinks positionSize when broker reports a smaller balance than persisted`
- `does not reconcile a fresh strategy that has not yet attached`

Full suite: 19/19 pass in this file, 180/180 across `trading-strategies`.

## Related

The companion guard in `TradingSession` (refuse to forward `qty <= 0` to the broker, regardless of strategy) ships in a separate PR so each fix is reviewable in isolation.

### DIFF
--- a/packages/trading-strategies/src/strategy-trailing-stop/TrailingStopStrategy.ts
+++ b/packages/trading-strategies/src/strategy-trailing-stop/TrailingStopStrategy.ts
@@ -164,19 +164,18 @@ export class TrailingStopStrategy extends Strategy {
 
     /*
      * Post-restore reconciliation. Runs at most once per process, on the first candle
-     * where the strategy is already attached (peakPrice !== '0'). If the broker reports
-     * a smaller base balance than the strategy's persisted positionSize, the position
-     * was reduced or fully closed outside the strategy — usually because a previous
-     * fill never propagated through `onFill` (e.g. unclean shutdown between fill arrival
-     * and state persistence). Without this check, the strategy would keep re-emitting a
-     * sell-all every minute against a phantom position; `TradingSession` would then
-     * resolve the size to 0 and either drop the order (post-Bug-B fix) or send qty=0
-     * to the broker (pre-fix) — either way, log spam without forward progress.
+     * where the strategy is already attached (peakPrice !== '0'). The broker is the source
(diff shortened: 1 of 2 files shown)


packages/trading-strategies/src/strategy-trailing-stop/TrailingStopStrategy.ts
```diff
@@ -153,6 +162,40 @@ export class TrailingStopStrategy extends Strategy {
       return undefined;
     }
 
+    /*
```
bennycode, line 165: @claude in this case, can we take the qty received from the broker (source of truth) as our new available balance?

PR #1097, opened 2026-05-29
fix(exchange): refuse to forward orders with non-positive resolved size

bennycode:
## Summary

Companion to #1096. When a strategy emits `AllAvailableAmount` advice and the relevant balance is zero, `TradingSession.#resolveOrderSize` returns `0`, and the session forwards it to the broker. Alpaca rejects with HTTP 422 `qty must be > 0` — once a minute, indefinitely, while the level-50 error logs pile up.

The existing `< minSize` check doesn't catch this because Alpaca sets `base_min_size = "0"` for US equities — so `Big(0).lt("0")` is `false` and the zero-sized order slips through.

Adds an explicit `size.lte(0)` guard immediately after `#resolveOrderSize`. The session emits an `'error'` event with a message naming the side and `amountIn`, then returns — no broker call, no HTTP 422 from Alpaca.

The strategy still owns the underlying state-drift bug (handled in #1096); this is defense in depth so any other strategy that ever resolves to zero gets the same protection without depending on the Alpaca-specific min-size value.

## Test coverage

`packages/exchange/src/trader/TradingSession.test.ts`:
- `refuses to place an order when AllAvailableAmount resolves to zero`

Full suite: 19/19 pass in this file, 88/88 across `exchange`.

### DIFF
--- a/packages/exchange/src/trader/TradingSession.test.ts
+++ b/packages/exchange/src/trader/TradingSession.test.ts
@@ -13,6 +13,7 @@ import {
   type PendingMarketOrder,
   type TradingRules,
 } from '../broker/Broker.js';
+import {NonPositiveOrderSizeError} from './TradingSessionErrors.js';
 import {AllAvailableAmount} from './TradingSessionTypes.js';
 import type {OrderAdvice, TradingSessionStrategy} from './TradingSessionTypes.js';
 
@@ -472,7 +473,11 @@ describe.sequential('TradingSession', () => {
       exchange.emit('candle-topic-1', sampleCandle);
       await vi.waitFor(() => expect(onError).toHaveBeenCalledTimes(1));
 
-      expect(onError.mock.calls[0][0].message).toContain('non-positive size');
+      const error = onError.mock.calls[0][0];
+      expect(error).toBeInstanceOf(NonPositiveOrderSizeError);
+      expect(error.side).toBe(OrderSide.SELL);
+      expect(error.amountIn).toBe('base');
+      expect(error.size).toBe('0');
       expect(exchange.placeMarketOrder).not.toHaveBeenCalled();
       expect(exchange.placeLimitOrder).not.toHaveBeenCalled();
     });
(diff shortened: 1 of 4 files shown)


packages/exchange/src/trader/TradingSession.ts
```diff
@@ -178,6 +178,26 @@ export class TradingSession extends EventEmitter<TradingSessionEventMap> {
...
+     * equities — set `base_min_size` to "0", so the `< minSize` check below wouldn't catch
+     * this and the order would reach the exchange only to be rejected with 422 "qty must be
+     * > 0" once a minute. Skip the call, surface the situation, and let upstream fix the
+     * strategy state.
+     */
+    if (size.lte(0)) {
+      this.emit(
+        'error',
+        new Error(
+          `Refusing to place ${advice.side} order with non-positive size "${size.toFixed()}" ` +
```
bennycode: @claude put variables like `${advice.side}` in quotes `"`.

packages/exchange/src/trader/TradingSession.ts
```diff
@@ -178,6 +178,26 @@ export class TradingSession extends EventEmitter<TradingSessionEventMap> {
       return;
     }
 
+    /*
+     * Defense in depth: a strategy can legitimately request `AllAvailableAmount`, but if the
```
bennycode: @claude remove this big comment

packages/exchange/src/trader/TradingSession.test.ts
```diff
@@ -449,6 +449,34 @@ describe.sequential('TradingSession', () => {
...
+
+      const onError = vi.fn();
+      session.on('error', onError);
+
+      await session.start();
+
+      exchange.emit('candle-topic-1', sampleCandle);
+      await vi.waitFor(() => expect(onError).toHaveBeenCalledTimes(1));
+
+      expect(onError.mock.calls[0][0].message).toContain('non-positive size');
```
bennycode: todo: check for error type and not the error message itself

PR #1117, opened 2026-06-01
style: forbid redundant primitive return types across all packages

bennycode:
## Summary

Adds a lint guardrail (and cleanup) that enforces the workspace's existing "prefer inference" convention for **primitive** return types.

A `no-restricted-syntax` rule is added to every package's `eslint.config.mjs`:

```js
'no-restricted-syntax': ['error', {
  message: 'Drop the explicit primitive return type and rely on inference.',
  selector: ':matches(FunctionDeclaration, FunctionExpression, ArrowFunctionExpression) > TSTypeAnnotation > :matches(TSBooleanKeyword, TSNumberKeyword, TSStringKeyword, TSBigIntKeyword)',
}]
```

It flags explicit `: boolean | number | string | bigint` return annotations **only on functions with a body**, where the type is trivially inferred. It deliberately does **not** touch:
- abstract method signatures, interface/type members, overload declarations (no body — annotation is required)
- object / union / domain return types, including every zod `.parse()` method

## Cleanup

Removes **81** now-redundant annotations:

| Package | Removed |
|---|---|
| trading-signals | 35 |
| exchange | 17 |
| trading-strategies | 17 |
| messaging | 12 |

## Safety

- No `eslint-disable` exceptions were needed (no recursion/edge cases required keeping an annotation).
- **No public API / `.d.ts` change**: every affected exported member re-infers to the exact same primitive — verified by emitting and diffing declarations (notably for the published `trading-signals` library). No literal-union or narrowing surprises.
- `tsc --noEmit` clean and full test suites pass in all four packages.

## Why a rule, not just convention

The repo already documents "prefer inference" for return types; this makes it enforceable for the trivially-redundant primitive case without forcing annotations off genuinely useful domain/object return types.


packages/trading-strategies/eslint.config.mjs
```diff
@@ -18,6 +18,14 @@ export default defineConfig({
     // companion pattern as an enum replacement. `no-redeclare` misfires on the shared
     // value/type name even though TypeScript allows it; renaming would break public APIs.
     '@typescript-eslint/no-redeclare': 'off',
+    'no-restricted-syntax': [
```
bennycode: todo: why do we have this rule 3x in the repo?

bennycode:
todo: make sure primites are also not put a explicit return type when the return type is a Promise like `Promise<void>`.

PR #1121, opened 2026-06-08
feat(trading-strategies): volatility-sized dynamic trailing stop

bennycode:
## Summary

Adds a trailing-stop strategy whose trail width is sized from the instrument's own volatility (**ATR%**) rather than a hardcoded percentage, so one config adapts across instruments **and** timeframes. The investigation that motivated this (a real STX whipsaw) is captured as regression tests, and a suitability report makes "is a trailing stop the right tool here?" a one-call check.

## `trading-strategies`
- **`DynamicTrailStrategy`** (ratcheting) and **`ChandelierTrailStrategy`** (fully adaptive) over a shared `AbstractDynamicTrailStrategy`.
- Optional **daily-ATR aggregation** (`atrIntervalMillis`) with intraday breach checks, plus an **`init()` warm-up** that pre-seeds the ATR from historical candles so there's no cold-start blind spot.
- Utils: `trailStop`, `AtrPercent`, `atrUnits` (percent ⇄ ATR multiples + whippy/balanced/loose), `TrendFilter`, `MarketRegimeFilter`.
- **`assessTrailSuitability`** — backtests an instrument and reports opportunity-cost vs protection (`protective-fit` / `rode-through` / `opportunity-cost` / `whipsaw-prone`).
- STX and MSFT regression tests pinned to real fills.

## `@typedtrader/candles` (new package)
- Labeled market-condition candle datasets (uptrend / downtrend / sideways, ticker variants, STX, URTH) as `json/` plus a typed loader. `trading-signals-docs` now re-exports from it instead of holding its own copies.

## `exchange`
- Import the missing `node:assert` in `BrokerMock` — it previously "worked" only because vitest provides `assert` as a global, and threw under any other runtime.
- `HistoricalCandleProvider` + optional strategy `init()` warm-up hook, wired into `TradingSession.start()`; `getCandles` added to `TradingSessionBroker`.

## `trading-signals`
- Export the `MovingAverage` base class.

## Notes
- New cross-package deps mean release ordering matters: `trading-signals` and `@typedtrader/candles` (and the `exchange` interface additions) should publish before the dependents.
- `@typedtrader/candles` starts at `0.1.0`.


bennycode:
Split into a 6-PR dependency-ordered stack for easier review (merge bottom-up):

1. #1122 — fix(exchange): import node:assert in BrokerMock
2. #1123 — feat(trading-signals): export MovingAverage base class
3. #1124 — feat(exchange): strategy warm-up hook
4. #1125 — feat(candles): @typedtrader/candles fixtures package
5. #1126 — feat(trading-strategies): volatility trailing-stop utils
6. #1127 — feat(trading-strategies): DynamicTrailStrategy + suitability report

The top of the stack (#1127) is byte-identical to this branch.

PR #1124, opened 2026-06-09
feat(exchange): strategy warm-up hook from historical candles

bennycode:
### 📚 PR stack — #1121 split into 6 (review/merge bottom-up)

- #1122 — fix(exchange): import node:assert in BrokerMock
- #1123 — feat(trading-signals): export MovingAverage base class
- 👉 **#1124 — feat(exchange): strategy warm-up hook from historical candles** ← this PR
- #1125 — feat(candles): @typedtrader/candles fixtures package
- #1126 — feat(trading-strategies): volatility trailing-stop utils
- #1127 — feat(trading-strategies): DynamicTrailStrategy + suitability report

---

Lets a strategy **pre-seed its indicators from history** before the first live candle, instead of sitting cold (unprotected) through the warm-up period.

- `HistoricalCandleProvider` + optional `TradingSessionStrategy.init()`, invoked by `TradingSession.start()` **before** subscribing to live candles.
- `getCandles()` added to the `TradingSessionBroker` interface (`AlpacaBroker` already implements it); `BrokerMock` gains a seedable `getCandles` for tests.

Exchange suite: 98 tests pass (incl. a new wiring test that `start()` calls `init()` with a `getCandles`-backed provider).

_3/6 of the stack splitting #1121. Based on #1123._


packages/exchange/src/trader/TradingSession.ts
```diff
@@ -69,6 +69,14 @@ export class TradingSession extends EventEmitter<TradingSessionEventMap> {
       tradingRules,
     };
 
+    /*
```
bennycode: todo: remove comment because a strategy could do other things than just warming up its candles.

packages/exchange/src/trader/TradingSessionTypes.ts
```diff
@@ -16,7 +18,27 @@ import type {
 } from '../broker/Broker.js';
 import type {TradingPair} from '../broker/TradingPair.js';
 
+/**
+ * Read-only market-data context handed to a strategy's {@link TradingSessionStrategy.init}. It lets
+ * the strategy pull whatever history it needs to warm up (e.g. 300 hourly candles for an ATR) but
+ * deliberately exposes no order placement — warm-up reads data, it never trades. Both real brokers
+ * and standalone {@link MarketDataSource}s satisfy this shape.
+ */
+export interface MarketDataProvider {
```
bennycode: todo: i don't want an extra interface, can't we just put this inside of `MarketDataSource`?

packages/exchange/src/trader/TradingSessionTypes.ts
```diff
@@ -16,7 +18,27 @@ import type {
...
+  getCandles(pair: TradingPair, request: CandleImportRequest): Promise<Candle[]>;
+  getLatestCandle(pair: TradingPair, intervalInMillis: number): Promise<Candle>;
+  /** Fetch the most recent `count` candles of the given interval, oldest first. */
+  getRecentCandles(pair: TradingPair, count: number, intervalInMillis: number): Promise<Candle[]>;
+}
+
 export interface TradingSessionStrategy {
+  /**
```
bennycode: todo: remove comment because init strategy can be unrelated to warming up candles.

packages/exchange/src/broker/BrokerMock.ts
```diff
@@ -38,6 +40,20 @@ export abstract class BrokerMock extends Broker {
     this.#balances = config.balances;
   }
 
+  /** Seed the candles returned by {@link getCandles} (used to exercise strategy warm-up). */
+  setHistoricalCandles(candles: Candle[]) {
+    this.#historicalCandles = candles;
+  }
+
+  async getCandles(_pair: TradingPair, _request: CandleImportRequest): Promise<Candle[]> {
```
bennycode: todo: why do we need this? we have "getRecentCandles" and "Strategy.init" already.

packages/exchange/src/trader/TradingSession.ts
```diff
@@ -69,6 +69,14 @@ export class TradingSession extends EventEmitter<TradingSessionEventMap> {
...
+     * Let the strategy initialize before the first live candle arrives. It only gets read access to
```
bennycode: todo: remove comment

packages/candles/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  },
+  {
+    candles: urth2025Candles,
+    description: 'iShares MSCI World ETF (URTH) - full year 2025 (1h)',
+    id: 'urth-2025',
+    name: 'MSCI World ETF (2025)',
+  },
+];
+
+/** Look up a dataset by its `id` (e.g. `'uptrend-stx'`), or `undefined` if there is no match. */
```
bennycode: todo: remove comment

packages/exchange/src/broker/alpaca/AlpacaBroker.ts
```diff
@@ -131,6 +131,10 @@ export class AlpacaBroker extends Broker implements MarketDataSource {
     return this.#marketData.getLatestCandle(pair, intervalInMillis);
   }
 
+  async getRecentCandles(pair: TradingPair, count: number, intervalInMillis: number): Promise<Candle[]> {
```
bennycode, line 134: ask: why do we have to expose that from the broker? we already have it on the market data provider.

packages/exchange/src/broker/MarketDataSource.ts
```diff
@@ -16,6 +16,53 @@ import type {TradingPair} from './TradingPair.js';
 export abstract class MarketDataSource extends EventEmitter {
   abstract getCandles(pair: TradingPair, request: CandleImportRequest): Promise<Candle[]>;
   abstract getLatestCandle(pair: TradingPair, intervalInMillis: number): Promise<Candle>;
+
+  /**
```
bennycode, line 20: todo: shrink comment. the first sentence is enough.

packages/exchange/src/trader/TradingSession.ts
```diff
@@ -69,6 +69,10 @@ export class TradingSession extends EventEmitter<TradingSessionEventMap> {
       tradingRules,
     };
 
+    if (this.#strategy.init) {
```
bennycode: ask: cant we use `?` operator here?

packages/exchange/src/broker/alpaca/AlpacaBroker.ts
bennycode: It's exposed because the session hands the **broker itself** to `strategy.init(market, pair)` as the read-only data context, and `AlpacaBroker implements MarketDataSource` — so it has to surface `getRecentCandles` (here it just delegates to the inner `#marketData`).

It can't live *only* on the market-data provider because not every broker has a separate one: the mock/backtest brokers (`BrokerMock`) **are** their own data source and implement `getRecentCandles` directly, so there'd be no inner provider to hand out instead. Keeping the method on the broker is the one uniform shape that works for both real (delegating) and mock (self-sourcing) brokers.

If you'd rather, I can expose the underlying `MarketDataSource` and pass that to `init` instead — but the mocks would then need a wrapper, which is why I kept the one-line delegation.

PR #1126, opened 2026-06-09
feat(trading-strategies): volatility trailing-stop utils

bennycode:
### 📚 PR stack — #1121 split into 6 (review/merge bottom-up)

- #1122 — fix(exchange): import node:assert in BrokerMock
- #1123 — feat(trading-signals): export MovingAverage base class
- #1124 — feat(exchange): strategy warm-up hook from historical candles
- #1125 — feat(candles): @typedtrader/candles fixtures package
- 👉 **#1126 — feat(trading-strategies): volatility trailing-stop utils** ← this PR
- #1127 — feat(trading-strategies): DynamicTrailStrategy + suitability report

---

Reusable building blocks for sizing a trailing stop to an instrument's own volatility (consumed by the strategy in the next PR):

- **`trailStop`** — percent + ATR trail-price helpers (returns the looser of the two).
- **`AtrPercent`** — ATR as a percent of price (the "usual % move per bar").
- **`atrUnits`** — convert between a percentage stop and ATR multiples, plus a `whippy`/`balanced`/`loose` classification.
- **`TrendFilter`** — price-vs-moving-average gate (uses the `MovingAverage` base from #1123).
- **`MarketRegimeFilter`** — risk-on/off gate driven by an index series.

35 unit tests; typecheck clean.

_5/6 of the stack splitting #1121. Based on #1125._

PR #1131, opened 2026-06-19
feat(trading-strategies): add AtrTrailStrategy (ATR-sized trailing stop)

bennycode:
A simpler alternative to `DynamicTrailStrategy`: a trailing stop whose width is sized **once** from the stock's own volatility, then left alone.

## What it does
- On `init(market, pair)` it pulls recent history via `getRecentCandles`, measures **ATR%** (`AtrPercent`), and **freezes** the trail width at `atrMultiple × ATR%`.
- From then on it's an ordinary percentage trailing stop: ratchet the peak up, exit (market sell) when `close` drops to `peak × (1 − trailPct/100)`.
- No live ATR feed and no per-candle re-sizing — that's the whole point of being simpler than `DynamicTrailStrategy`.

## Why
Sizes a volatility-aware stop distance for the specific instrument (wide for a volatile name like STX, tight for a calm one) without hand-tuning a percentage — which is exactly what avoids the too-tight fixed-% whipsaw.

## Usage
```ts
new AtrTrailStrategy({atrInterval: 14, atrMultiple: '2', atrIntervalMillis: 86_400_000});
```

## Notes
- Exit-only; attaches to whatever base balance exists (incl. across a restart).
- Reuses the merged building blocks from #1124 (warm-up hook) and #1126 (`AtrPercent`).
- If history is too short to size the ATR, it holds without a stop and says so.
- Tradeoff vs `DynamicTrailStrategy`: it won't adapt to a mid-trade volatility regime shift — acceptable for "size sensibly from recent history, then just trail."

7 unit tests cover sizing, attach, ratchet, breach/exit, the no-history and no-init cases.


packages/trading-strategies/src/strategy-atr-trail/AtrTrailStrategy.ts
```diff
@@ -0,0 +1,266 @@
...
+  PendingOrder,
+  TradingPair,
+  TradingSessionState,
+} from '@typedtrader/exchange';
+import {MarketType} from '../strategy/MarketType.js';
+import {Strategy} from '../strategy/Strategy.js';
+import {AtrPercent} from '../util/AtrPercent.js';
+import {positiveNumberString} from '../util/validators.js';
+
+const ONE_DAY_IN_MS = 86_400_000;
```
bennycode: no need for this, we can use the `ms` library (`ms('1d')`).

packages/trading-strategies/src/strategy-atr-trail/AtrTrailStrategy.ts
```diff
@@ -0,0 +1,266 @@
...
+    super.restoreState(validated);
+    const restored = this.#state;
+    restored.exited = validated.exited;
+    restored.peakPrice = validated.peakPrice;
+    restored.stopPrice = validated.stopPrice;
+    restored.trailDownPct = validated.trailDownPct;
+  }
+}
+
+function isAtrTrailState(value: unknown): value is AtrTrailState {
```
bennycode: this is a huge function, why do we need to check each value in the obj?

bennycode: The validation is there because `restoreState` receives arbitrary persisted JSON from disk (a snapshot that could be stale, partial, or hand-edited) — a malformed state would silently strand the strategy in an inconsistent runtime (e.g. a peak with no matching stop). So I kept the check but replaced the ~35-line hand-rolled type-guard with a small zod schema and `safeParse` (fallback to defaults), which is both shorter and matches the codebase's zod convention. Added tests for the valid-restore and malformed-fallback paths.

bennycode:
Wired the `AtrTrail` indicator (#1236) into this strategy:

- Trail-width sizing (`trailPct = atrMultiple × ATR%`) is now delegated to `AtrTrail` from `trading-signals` instead of the local `AtrPercent` util. The indicator runs in ROLLING mode internally even when this strategy's `rolling` config is off — its FROZEN mode locks the width at the *first* stable candle, but the strategy sizes from the *full* warm-up history, so freezing stays the strategy's job (it simply stops feeding candles after `init`).
- The peak/stop ratchet stays in the strategy: it's position-scoped (highest high since attach, persisted across restarts), while the indicator's peak spans the whole feed.
- **Note:** this branch temporarily contains #1236's commits (merged in, since #1236 isn't on main yet). Once #1236 merges, a main merge-back collapses them out of this PR's diff.
- Fixed a merge-integration break from main's `trader/` extraction: `AllAvailableAmount`, `OrderAdvice` and `TradingSessionState` imports from `@typedtrader/exchange` resolved to `undefined` at runtime, which broke the sentinel check in `AdviceExecutor` and silently disabled the STX regression tests. All imports now come from `../trader/index.js`; all 14 strategy tests (incl. the 3 STX regression tests) pass again.

PR #1178, opened 2026-07-05
fix(@typedtrader/exchange): send client order ids with order placements

bennycode:
## Summary

Eliminates a duplicate-order risk from blind retries of order placement.

**The scenario:** both broker API clients wire `axios-retry` with a method-blind retry condition (Alpaca: 20 retries; Trading212: `retries: Infinity`). When an order-placement POST times out or returns a 5xx, the order may still have been **accepted server-side** — the failure only means the client never saw the response. A blind retry then re-submits the same order and can silently open a duplicate real-money position. Order placement is not idempotent, so it must never be auto-retried.

Changes:

- **Alpaca** (`AlpacaBroker.placeOrder`): every placement now generates a `client_order_id` via `crypto.randomUUID()` and sends it in the `POST /v2/orders` body. Alpaca deduplicates on this id, so a manual re-submission with the same id is safe, and operators can reconcile whether a failed request actually landed. ([docs](https://docs.alpaca.markets/docs/orders-at-alpaca#client-order-id))
- **Alpaca** (`shouldRetryAlpacaRequest`): `POST /v2/orders` is never retried. GETs/DELETEs (including reads of `/v2/orders` and order cancellations) keep the existing retry behavior, and the PDT (40310100) / no-short (40310000) business-error filtering is unchanged.
- **Trading212** (`shouldRetryTrading212Request`, extracted + exported): POSTs to `/api/v0/equity/orders/market` and `/api/v0/equity/orders/limit` are never retried — Trading212's public API has no idempotency-key support, so a duplicate would be unrecoverable. All other requests keep today's behavior, including the per-endpoint rate-limit delays in `getRetryDelay`.

A failed order POST now surfaces immediately as an error to the caller (TradingSession already converts it to an `error` event; that path is untouched). Public package types are unchanged — neither API class is exported from `src/index.ts`, so `postOrder` could gain a required `client_order_id` param to enforce the invariant at compile time.

## Test plan

- [x] `AlpacaAPI.test.ts`: order POST not retried on 503/429/network errors (incl. uppercase method); GET/DELETE on the same endpoint still retried; 429/503/EAI_AGAIN on other requests still retried; PDT/no-short rejections still not retried
- [x] `Trading212API.test.ts` (new): market/limit order POSTs not retried on 503/network errors; GET/DELETE order endpoints and other requests keep retry behavior; HTTP 400 not retried
- [x] `AlpacaBroker.test.ts`: `placeOrder` passes a `client_order_id` matching the UUIDv4 shape on the mocked API call; unique across two placements
- [x] `packages/exchange`: `npm test` — 9 files, 103 tests passed
- [x] `packages/exchange`: `npm run lint` — clean
- [x] `npx lerna run dist --scope trading-strategies --include-dependencies` — 3 projects built


packages/exchange/src/broker/alpaca/api/AlpacaAPI.ts
```diff
@@ -154,8 +170,12 @@ export class AlpacaAPI {
     await this.#tradingClient.delete(`/v2/orders/${orderId}`);
   }
 
-  /** @see https://docs.alpaca.markets/reference/postorder */
+  /**
+   * @see https://docs.alpaca.markets/reference/postorder
```
bennycode, line 158: todo: link to https://docs.alpaca.markets/us/docs/working-with-orders#using-client-order-ids

packages/exchange/src/broker/alpaca/AlpacaBroker.test.ts
```diff
@@ -220,6 +224,8 @@ describe('AlpacaBroker', {concurrent: false}, () => {
   });
 
   describe('placeOrder', () => {
+    const UUID_PATTERN = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
```
bennycode: todo: validate with `z.uuid` here

packages/exchange/src/broker/alpaca/AlpacaBroker.ts
```diff
@@ -402,6 +403,13 @@ export class AlpacaBroker extends Broker implements MarketDataSource {
       time_in_force: string;
       type: string;
     } = {
+      /*
+       * Order placement is never auto-retried (a blind retry can duplicate a position), so a
+       * failed POST needs a client-side id: it lets operators reconcile whether the original
+       * request landed and makes any manual re-submission with the same id deduplicated by Alpaca.
+       * @see https://docs.alpaca.markets/docs/orders-at-alpaca#client-order-id
+       */
+      client_order_id: randomUUID(),
```
bennycode, line 413: this is a very good example, add this kind of reconciliation behaviour as best pratices to to https://github.com/bennycode/trading-signals/blob/main/packages/exchange/BROKER_TEMPLATE.md

packages/exchange/src/broker/trading212/api/Trading212API.ts
```diff
@@ -84,9 +105,7 @@ export class Trading212API {
 
     axiosRetry(this.#httpClient, {
       retries: Infinity,
-      retryCondition: (error: AxiosError) => {
-        return axiosRetry.isRetryableError(error);
-      },
+      retryCondition: shouldRetryTrading212Request,
```
bennycode: leave Trading212 out of scope, do order submission reconciliation only for Alpaca exchange first

packages/exchange/src/broker/alpaca/api/AlpacaAPI.ts
```diff
@@ -36,7 +36,23 @@ function isNotAllowedToShort(error: AxiosError) {
...
+ * whether the original request landed.
+ *
+ * @see https://docs.alpaca.markets/docs/orders-at-alpaca#client-order-id
+ */
+function isOrderPlacement(error: AxiosError) {
+  return error.config?.method?.toLowerCase() === 'post' && error.config.url === '/v2/orders';
+}
+
 export function shouldRetryAlpacaRequest(error: AxiosError) {
+  if (isOrderPlacement(error)) {
```
bennycode: todo: remove excluding order retry based on url. we are now sending the client order id which should be good enough for reconciliation

PR #1185, opened 2026-07-09
docs(trading-strategies): document strategy contract, runtimes, and data access

bennycode:
## What

The `trading-strategies` README explained trading *domain knowledge* well but was thin on the actual *programming contract*. This adds the missing architectural docs and fixes some stale/incorrect examples.

## Changes

- **New "The Strategy Contract" section** — the candle-in / advice-out model that wasn't documented anywhere: `processCandle(candle, state)` as the one method you implement, the `TradingSessionState` input, the real `OrderAdvice` output union, and the full lifecycle-hook table (`init`/`onFill`/`onOrderFilled`/`onMessage`). Also clarifies `onCandle` is a template method — you override `processCandle`.
- **New "Running a Strategy" section** — documents the two runtimes (`BacktestExecutor` historical vs `TradingSession` live) and the key fact that both route advice through the **same `AdviceExecutor`**, so a backtest is a faithful dry run of the live path. Includes runnable examples for each.
- **New "Strategies Are Just Async Code" section** — makes explicit that `processCandle`/`init` are async by design and may `await` any data source (DB, news, analyst/stock reports, an MCP such as TipRanks), not only `trading-signals` indicators. Cites `MeanReversionStrategy`'s `fetchCandles` and `init(market)` as existing proof, with a perf caveat to keep the per-candle path cheap.
- **Rewrote "Strategy Signals"** — the previous `BUY_MARKET`/`SELL_LIMIT`/`NONE` enum did not exist in the code; replaced with the real `OrderAdvice` shapes.
- **Fixed Usage imports** — `OrderAdvice`/`AllAvailableAmount` come from `trading-strategies` (were imported from `@typedtrader/exchange`); `OrderSide`/`OrderType` are the real names (were the non-existent `ExchangeOrderSide`/`ExchangeOrderType`).

## Notes

Docs-only change. All type names and imports were verified against the current source (`Strategy.ts`, `TradingSessionTypes.ts`, `BacktestExecutor.ts`, `MeanReversionStrategy.ts`).

### DIFF
--- a/packages/trading-strategies/README.md
+++ b/packages/trading-strategies/README.md
@@ -22,132 +22,47 @@ The "trading-strategies" library provides a TypeScript implementation for common
 npm install trading-strategies
 ```
 
-## Usage
+## How a Strategy Works
 
-```ts
-import {Strategy, AllAvailableAmount} from 'trading-strategies';
-import type {OrderAdvice} from 'trading-strategies';
-import {OrderSide, OrderType} from '@typedtrader/exchange';
-```
+A strategy watches the market one candle at a time and decides what to do next: **buy, sell, or wait.** You write that decision; the library handles the rest — turning it into a correctly sized order and keeping track of balances, fees, and each exchange's trading rules for you.
 
-## The Strategy Contract
+On every new candle your strategy sees the latest price data and a snapshot of your account, then returns one of:
 
-A strategy is **candles in, advice out**. You implement a single method; the runtime does the rest.
+- **Buy** — at the current price (market order) or at a price you set (limit order)
+- **Sell** — at the current price or at a price you set
+- **Nothing** — sit this one out
 
-```ts
-protected abstract processCandle(
-  candle: OneMinuteBatchedCandle,
-  state: TradingSessionState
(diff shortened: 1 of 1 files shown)


packages/trading-strategies/README.md
```diff
@@ -22,14 +22,48 @@ The "trading-strategies" library provides a TypeScript implementation for common
 npm install trading-strategies
 ```
 
-## Usage
+## How a Strategy Works
+
+A strategy watches the market one candle at a time and decides what to do next: **buy, sell, or wait.** You write that decision; the library handles the rest — turning it into a correctly sized order and keeping track of balances, fees, and each exchange's trading rules for you.
```
bennycode, line 27: @claude don't use this `-` AI writing style

packages/trading-strategies/README.md
```diff
@@ -22,14 +22,48 @@ The "trading-strategies" library provides a TypeScript implementation for common
...
+
+On every new candle your strategy sees the latest price data and a snapshot of your account, then returns one of:
```
bennycode, line 29: For this showcase an example code that returns an advice.

PR #1188, opened 2026-07-11
feat: fee-aware round-trip profit helper + SMA crossover that won't sell at a loss

bennycode:
## Why

The SMA crossover bleeds money on trading fees: it sells on every bearish cross, including round trips where the sell price barely differs from the buy price — a "win" on raw price that a maker+taker fee round trip quietly turns into a loss. There was no helper anywhere to check whether a sale actually nets a profit after fees.

## What

Four commits, bottom-up:

### 1. `feat(exchange)` — round-trip profit helper
`calculateRoundTripProfit({entryPrice, exitPrice, quantity, entryFeeRate, exitFeeRate})` → `{ netProfit, netProfitPercent, entryCost, exitProceeds, entryFee, exitFee, breakEvenPrice, isProfitable }`, plus `isProfitableAfterFees(...)`. Fee-inclusive on **both** legs; also reports the **break-even sell price**. Pure util in `@typedtrader/exchange`, 10 tests.

### 2. `feat(trading-strategies)` — `FeeAwareSmaCrossoverStrategy`
Extends `SmaCrossoverStrategy` and gates only the SELL: on a bearish cross it forwards the SELL only when the round trip nets a profit after the taker fee on both legs — otherwise it **holds**. Buys pass through unchanged. Registered + exported.

### 3. `feat(trading-signals-docs)` — backtest page
Adds **"SMA Crossover (Fee-Aware)"** to the backtest UI to compare head-to-head with the plain crossover.

### 4. `refactor(trading-strategies)` — position tracking on the base `Strategy`
Moves fill-driven position tracking into the root `Strategy`: a base `onFill` maintains `positionSize` and a cost-basis-averaged `averageEntryPrice`, persisted under a reserved state key. Any strategy can now measure an exit against its entry without its own bookkeeping.
- `FeeAwareSmaCrossoverStrategy` drops its own last-buy tracking and reads `averageEntryPrice` from the base (correct under partial fills, not just all-in/out).
- `ProtectedStrategy` and `TrailingStopStrategy` now `override onFill` with a `super.onFill()` call so base tracking stays accurate (they keep their own state too, for now — a later cleanup could collapse the duplication).
- Adds `Strategy.test.ts` covering buys, partial sells, full exit, and persistence.

## Trade-off (documented in the class)

Skipping unprofitable exits means it can **hold through a downtrend** rather than take a small loss. Pair with a stop-loss for a hard floor. This is a fee-bleed fix, not a risk-management change.

## Verification

`exchange`: 10 helper tests. `trading-strategies`: full suite **269** green (incl. new base tracking). Docs `tsc` clean. Lint clean across all packages.

### DIFF
--- a/packages/trading-signals-docs/utils/strategySchemas.ts
+++ b/packages/trading-signals-docs/utils/strategySchemas.ts
@@ -1,10 +1,10 @@
 import {z} from 'zod';
 import type {Candle} from '@typedtrader/exchange';
-import {BuyOnceSchema, BuyBelowSellAboveSchema, CoinFlipSchema, MultiIndicatorConfluenceSchema, ScalpSchema, MeanReversionSchema, SmaCrossoverSchema, ProtectedStrategySchema, TrailingStopSchema, suggestScalpOffset} from 'trading-strategies';
+import {BuyOnceSchema, BuyBelowSellAboveSchema, CoinFlipSchema, MultiIndicatorConfluenceSchema, ScalpSchema, MeanReversionSchema, SmaCrossoverSchema, FeeAwareSmaCrossoverSchema, ProtectedStrategySchema, TrailingStopSchema, suggestScalpOffset} from 'trading-strategies';
 
-export {BuyOnceSchema, BuyBelowSellAboveSchema, CoinFlipSchema, MultiIndicatorConfluenceSchema, ScalpSchema, MeanReversionSchema, SmaCrossoverSchema, ProtectedStrategySchema, TrailingStopSchema};
+export {BuyOnceSchema, BuyBelowSellAboveSchema, CoinFlipSchema, MultiIndicatorConfluenceSchema, ScalpSchema, MeanReversionSchema, SmaCrossoverSchema, FeeAwareSmaCrossoverSchema, ProtectedStrategySchema, TrailingStopSchema};
 
-export type StrategyId = 'buy-and-hold' | 'buy-once' | 'buy-below-sell-above' | 'coin-flip' | 'multi-indicator-confluence' | 'scalp' | 'mean-reversion' | 'sma-crossover' | 'protection-only' | 'trailing-stop';
+export type StrategyId = 'buy-and-hold' | 'buy-once' | 'buy-below-sell-above' | 'coin-flip' | 'multi-indicator-confluence' | 'scalp' | 'mean-reversion' | 'sma-crossover' | 'fee-aware-sma-crossover' | 'protection-only' | 'trailing-stop';
 
 export interface StrategyDefinition {
   id: StrategyId;
@@ -128,6 +128,14 @@ export const strategyDefinitions: StrategyDefinition[] = [
     description:
       'Crosses a fast SMA against a slow SMA — each with its own period and timeframe (e.g. SMA5 on 1m bars vs SMA10 on 2m bars). Buys when the fast SMA crosses above the slow one and sells when it crosses below (market orders).',
     schema: SmaCrossoverSchema,
+    getDefaultConfig: () => ({fastPeriod: 5, fastTimeframe: '1m', slowPeriod: 10, slowTimeframe: '2m'}),
+  },
+  {
+    id: 'fee-aware-sma-crossover',
+    name: 'SMA Crossover (Fee-Aware)',
+    description:
+      'Same as SMA Crossover, but it never sells at a loss: on a bearish cross it only sells when the round trip — last buy price vs current price, with the taker fee on both legs — nets a profit. Otherwise it holds. Cuts the fee bleed from whipsaw exits (trade-off: it can hold through a downtrend, so add a stop-loss for a hard floor).',
+    schema: FeeAwareSmaCrossoverSchema,
     getDefaultConfig: () => ({fastPeriod: 5, fastTimeframe: '1m', slowPeriod: 10, slowTimeframe: '2m'}),
   },
   {
--- a/packages/trading-strategies/src/strategy/Strategy.ts
+++ b/packages/trading-strategies/src/strategy/Strategy.ts
@@ -93,14 +93,30 @@ export abstract class Strategy implements TradingSessionStrategy {
 
   async onCandle(candle: OneMinuteBatchedCandle, state: TradingSessionState): Promise<OrderAdvice | void> {
     this.lastBatchedCandle = candle;
-    const advice = await this.processCandle(candle, state);
+    const advice = await this.decideAdvice(candle, state);
     this.latestAdvice = advice ? advice : undefined;
     return advice;
   }
 
+  /**
+   * Turns a candle into advice. Defaults to `processCandle`. A base that needs to gate candle
+   * handling (like ProtectedStrategy while a kill switch is active) overrides this instead of
+   * `onCandle`, so the candle and advice bookkeeping above is never duplicated or forgotten.
+   */
+  protected async decideAdvice(
+    candle: OneMinuteBatchedCandle,
+    state: TradingSessionState
+  ): Promise<OrderAdvice | void> {
+    return this.processCandle(candle, state);
+  }
+
   restoreState(persisted: Record<string, unknown>): void {
     this.#_state = persisted;
+    this.hydrateState(persisted);
   }
+
+  /** Override to copy persisted fields into the live state on restore. The base has already stored the snapshot. */
+  protected hydrateState(_persisted: Record<string, unknown>): void {}
 
   /**
    * Remembers the last buy and sell price, then lets subclasses react by overriding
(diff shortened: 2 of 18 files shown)


packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -1,8 +1,23 @@
 import Big from 'big.js';
-import type {OneMinuteBatchedCandle} from '@typedtrader/exchange';
+import {OrderSide} from '@typedtrader/exchange';
+import type {Fill, OneMinuteBatchedCandle} from '@typedtrader/exchange';
 import type {OrderAdvice, TradingSessionState, TradingSessionStrategy} from '../trader/index.js';
 import type {MarketType} from './MarketType.js';
 
+const LAST_FILLS_STATE_KEY = '__lastFills__';
+
+/** Most recent fill price per side, persisted alongside the strategy's own state under a reserved key. */
```
bennycode: @claude comment is repetitive

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -1,8 +1,23 @@
...
+type LastFills = {
+  /** Price of the most recent BUY fill, as a Big string. `null` until the first buy. */
```
bennycode: @claude comment is repetitive

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -1,8 +1,23 @@
...
+  lastBuyPrice: string | null;
+  /** Price of the most recent SELL fill, as a Big string. `null` until the first sell. */
```
bennycode: @claude comment is repetitive

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -1,8 +1,23 @@
 import Big from 'big.js';
-import type {OneMinuteBatchedCandle} from '@typedtrader/exchange';
+import {OrderSide} from '@typedtrader/exchange';
+import type {Fill, OneMinuteBatchedCandle} from '@typedtrader/exchange';
 import type {OrderAdvice, TradingSessionState, TradingSessionStrategy} from '../trader/index.js';
 import type {MarketType} from './MarketType.js';
 
+const LAST_FILLS_STATE_KEY = '__lastFills__';
```
bennycode: @claude prefix this with the strategy name instead of doing `__`

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -82,6 +106,43 @@ export abstract class Strategy implements TradingSessionStrategy {
     this.#_state = persisted;
   }
 
+  /**
+   * Records the most recent BUY / SELL fill price, exposed as {@link lastBuyPrice} /
+   * {@link lastSellPrice}, so a strategy can measure an exit against its entry without
```
bennycode: @claude don't link classes and variables here all the time, it's prone when refactoring. Keep comments read-able for people and make them less technical.

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -82,6 +106,43 @@ export abstract class Strategy implements TradingSessionStrategy {
...
+   * bookkeeping of its own. Subclasses that override `onFill` MUST call
+   * `super.onFill(fill, state)` to keep this accurate.
```
bennycode: @claude can we enforce with an eslint rule that super is being called when overwriting?

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -82,6 +106,43 @@ export abstract class Strategy implements TradingSessionStrategy {
...
+   */
+  async onFill(fill: Fill, _state: TradingSessionState): Promise<void> {
+    const patch: Partial<LastFills> =
+      fill.side === OrderSide.BUY ? {lastBuyPrice: fill.price} : {lastSellPrice: fill.price};
+    this.#setLastFills(patch);
+  }
+
+  /** Price of the most recent BUY fill, or `undefined` if there hasn't been one. */
```
bennycode: @claude comment is repetitive

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -82,6 +106,43 @@ export abstract class Strategy implements TradingSessionStrategy {
...
+  get lastBuyPrice(): Big | undefined {
+    const price = this.#lastFills.lastBuyPrice;
+    return price === null ? undefined : new Big(price);
+  }
+
+  /** Price of the most recent SELL fill, or `undefined` if there hasn't been one. */
```
bennycode: @claude comment is repetitive

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -82,6 +106,43 @@ export abstract class Strategy implements TradingSessionStrategy {
...
+    return price === null ? undefined : new Big(price);
+  }
+
+  get #lastFills(): LastFills {
+    return this.getProxiedState<LastFillsContainer>()[LAST_FILLS_STATE_KEY];
+  }
+
+  #setLastFills(patch: Partial<LastFills>): void {
+    /*
+     * Reassigning the reserved top-level key trips the state Proxy's set trap, which fires
```
bennycode: @claude comment is repetitive

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Renamed to `strategyLastFills` (readable, name-prefixed) instead of the `__` convention — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Dropped the repetitive type/field comments on `LastFills` — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Dropped — the field name says it — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Dropped — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Removed the `{@link}` refs and rewrote the comments as plain prose — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Removed the getter doc (name + return type already say it) — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Removed — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Trimmed to a one-line note about why the whole key is replaced — 671c62c.

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: There's no built-in ESLint rule for 'must call super in an override' (only framework-specific plugins like React lifecycle rules), so it'd need a custom rule. Rather than enforce the super call, I removed the need for it: `onFill` now records the last buy/sell price and then calls a `processFill` hook that subclasses override — same template-method pattern as `onCandle`/`processCandle`. Base tracking always runs and there's nothing to forget. ProtectedStrategy/TrailingStop/Scalp now override `processFill`. Done in 671c62c.

bennycode:
Followed the template-method approach through to the other two 'override + must call `super`' spots (the same footgun behind the `onFill` eslint question):

- **onCandle** → base keeps the `lastBatchedCandle`/`latestAdvice` bookkeeping and delegates to a `decideAdvice` hook. ProtectedStrategy overrides `decideAdvice` for its kill-switch gate instead of reimplementing `onCandle`.
- **restoreState** → base stores the snapshot and calls a `hydrateState` hook. Nobody overrides `restoreState` anymore, so base state restoration can't be skipped; ProtectedStrategy subclasses chain `super.hydrateState` for guard-state restore.

Full suite 270 green, tsc + lint clean. cc77b1f

packages/trading-signals-docs/utils/strategySchemas.ts
```diff
@@ -130,6 +130,14 @@ export const strategyDefinitions: StrategyDefinition[] = [
     schema: SmaCrossoverSchema,
     getDefaultConfig: () => ({fastPeriod: 5, fastTimeframe: '1m', slowPeriod: 10, slowTimeframe: '2m'}),
```
bennycode, line 131: @claude let's rename "Timeframe" into "Resolution" - across the codebase for SmaCrossover strategy.

PR #1190, opened 2026-07-12
feat(trading-strategies): add declarative strategy graphs (GraphStrategy)

bennycode:
> **Stacked on #1189** — merge that first; GitHub will retarget this PR to `main` automatically.

## What

Strategies expressed as **data instead of code**: a versioned, Zod-validated JSON graph of typed building blocks, executed by `GraphStrategy` — an interpreter implementing `TradingSessionStrategy`. Graphs backtest and trade through the exact same executors as hand-written strategies; there is no separate toy runtime whose results could diverge.

## Building blocks (v1)

| Node | Role |
|---|---|
| `source:candle` | emits the incoming 1m candle |
| `batcher` | rolls candles up to a configurable timeframe; emits only on bar close |
| `field` | extracts open/high/low/close/volume/medianPrice |
| `indicator` | SMA / EMA / RSI with configurable period; silent until warmed up |
| `if` | compares two values; `onChange` trigger = crossover edge detection |
| `advice` | emits a market `OrderAdvice` when its trigger fires |

## Design highlights

- **Warm-up safety:** nodes emit only when stable (`getResultOrThrow()` strictly behind the `isStable` gate); held values between ticks make mixed timeframes structurally lookahead-free — a 5m indicator holds its last *closed* bar between batch closes.
- **Extension point:** `registerNodeType()` lets third parties add nodes (news signals, on-chain data). Evaluators support async `evaluate()` and an async `init()` hook (wired to the session/backtest init from #1189) for loading external datasets deterministically.
- **Validation:** node-addressable errors, cycle detection (Kahn), port type checking, single-writer inputs, sink requirement.

## Proof

The SMA crossover template graph produces **advice-for-advice identical output** to the hand-written `SmaCrossoverStrategy` across shared- and mixed-timeframe configs, and a **balance-identical backtest to 8 decimal places** (same trades, fees, P&L) via `BacktestExecutor`.

## Test plan

21 new tests: equivalence (3 configs + full backtest), validation errors, trigger semantics (`onChange` vs `always`), advice payloads, custom node registration incl. async evaluate/init. Full suite: 283 green, lint clean, build clean.

### DIFF
--- a/packages/trading-strategies/src/index.ts
+++ b/packages/trading-strategies/src/index.ts
@@ -1,5 +1,6 @@
 export * from './backtest/index.js';
 export * from './strategy/index.js';
+export * from './strategy-graph/index.js';
 export * from './trader/index.js';
 export {BuyOnceStrategy, BuyOnceSchema, type BuyOnceConfig} from './strategy-buy-once/BuyOnceStrategy.js';
 /** @deprecated Use {@link BuyOnceStrategy} without `buyAt` instead. */
--- a/packages/trading-strategies/src/strategy-graph/GraphSchema.ts
+++ b/packages/trading-strategies/src/strategy-graph/GraphSchema.ts
@@ -1,0 +1,31 @@
+import {z} from 'zod';
+
+/**
+ * A strategy expressed as data instead of code: the same JSON drives the visual editor,
+ * validation, sharing, and the {@link GraphStrategy} interpreter, so what users see on
+ * the canvas is exactly what runs in a backtest or live session.
+ */
+export const StrategyGraphSchema = z.object({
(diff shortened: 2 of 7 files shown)


bennycode:
Addressed all three review findings in dbdec89:
- **Deterministic evaluation order**: node ids and dependents are now sorted in the topological sort, so evaluation (and the first-advice-wins rule) depends only on graph structure — added a regression test racing two advice nodes under reversed JSON key/connection ordering
- **Empty-path Zod issues** (refinements) no longer produce a leading `: ` in config errors
- **Missing-sink error** reworded to match the actual check, since `registerNodeType()` allows custom sinks

PR #1191, opened 2026-07-12
feat(trading-signals-docs): add visual strategy builder

bennycode:
> **Stacked on #1190** (which stacks on #1189) — merge those first; GitHub retargets automatically.

## What

A visual, node-based strategy builder at `/graph-builder` — n8n/Scratch-style: drag building blocks, connect typed ports, configure them inline, and backtest. The canvas state **is** the graph JSON that `GraphStrategy` executes, so what users see is exactly what runs.

## Features

- **Palette from the engine's node registry** — nodes added via `registerNodeType()` appear automatically, no UI changes
- **Typed ports (Scratch guarantee):** mismatched kinds refuse to snap; occupied inputs reject a second writer; edge colors encode the value kind (candle/number/trigger)
- **Config forms inside nodes**, validated by the engine's Zod schemas
- **Live validation** by constructing the real interpreter on every change — the error shown is the error a backtest would throw, node-addressable
- **SMA crossover template**, clear canvas, per-node delete (✕ + Backspace), JSON share/import
- **Backtesting** via the existing `BacktestExecutor` with a buy-and-hold baseline, rendered by the existing `BacktestResults`

## Test plan

6 Playwright specs (class-based page object per repo e2e conventions), including the flagship: **build the SMA crossover from scratch** — palette clicks, config edits, 10 real mouse-dragged connections — and assert the backtest matches the template run exactly (trade count + ROI). Plus: template validity, results rendering, incomplete-graph rejection, type-mismatch refusal, node deletion. Full e2e suite 9/9 green; `next build` prerenders the page statically.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4360,6 +4360,55 @@
         "assertion-error": "^2.0.1"
       }
     },
+    "node_modules/@types/d3-color": {
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/@types/d3-color/-/d3-color-3.1.3.tgz",
+      "integrity": "sha512-iO90scth9WAbmgv7ogoq57O9YpKmFBbmoEoCHDB2xMBY0+/KVrqAaCDyCE16dUspeOvIxFFRI+0sEtqDqy2b4A==",
+      "license": "MIT"
+    },
+    "node_modules/@types/d3-drag": {
+      "version": "3.0.7",
+      "resolved": "https://registry.npmjs.org/@types/d3-drag/-/d3-drag-3.0.7.tgz",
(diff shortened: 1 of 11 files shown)


bennycode:
Addressed all three review findings in 0d7dd5b:
- **Number fields**: clearing the input now unsets the config key so the node's Zod schema default applies, instead of coercing `''` to `0`
- **`loadGraph()`** syncs `currentGraph` immediately, so validation and the run button no longer lag behind Clear/Template while the client-only canvas re-hydrates
- **Copy timer** is cleared on unmount

Also rebased onto the updated #1190 (deterministic topo sort). Full e2e suite re-run: green.

PR #1200, opened 2026-07-17
refactor(exchange): declare retry policy per endpoint with a @retry decorator

bennycode:
## Why

Both API clients configured retries through `axios-retry` interceptors that had to reverse-map failed request URLs/config back to policy. `Trading212API`'s `getRetryDelay` matched URLs via string comparison — a design that already produced one silent bug (strict `===` missing paginated `nextPagePath` URLs — see the apology comment it carried), and every new endpoint risked silently falling through to the default delay.

## What

**`@retry` method decorator** (TC39 standard mode, `packages/exchange/src/util/retry.ts`): each API method declares its own rate-limit policy right where the endpoint lives — no URL-matching table to keep in sync.

```ts
/** @see https://t212public-api-docs.redoc.ly/#operation/accountCash */
@retry({delayMs: 2_000})
async getAccountCash() { ... }
```

- Default retryable predicate matches the old `axiosRetry.isRetryableError` set: network errors, 429, 5xx (as `SimplifiedHttpError` statuses `0`/`429`/`5xx`, since `simplifyError` runs at the interceptor level).
- Per-endpoint delays carried over 1:1 from the old table.
- `getHistoryOrders` now pages through the decorated `getHistoryOrdersPage`, preserving **per-page** retry semantics (a decorator on the aggregate would restart pagination) and deleting the duplicated pagination loop.

**`AlpacaAPI` converted too** (second commit): Alpaca applies one rate-limit policy across all endpoints, so a single shared `retryAlpaca` decorator (20 retries, linear 1s backoff) replaces the axios-retry wiring on both axios clients. `shouldRetryAlpacaRequest` now operates on `SimplifiedHttpError` instead of raw `AxiosError`; the PDT (40310100) and short-selling (40310000) exclusions are preserved and covered by the existing test suite (+1 new case).

**Toolchain enablement:**

- Dropped vestigial `experimentalDecorators`/`emitDecoratorMetadata` from `tsconfig.lib.json` — no legacy decorators exist in the repo, and the flag forced legacy semantics that break standard-mode decorators.
- Added a scoped pre-transform plugin to `vitest.config.base.ts`: oxc (rolldown-vite/Vitest 4) cannot downlevel standard decorators yet, so decorator-bearing files route through esbuild (already present via tsx). Remove once oxc ships support.

## Behavior notes

- Order placement/cancel endpoints keep the old default backoff (`attempt × 1s`) — unchanged.
- `ECONNABORTED` timeouts would now retry where `axios-retry` excluded them; moot since neither client configures a timeout.
- `axios-retry` is fully removed as a dependency (third commit): the retry loop is extracted into a plain `withRetry()` that the decorator now adapts, and `isEarningsDay` — the last consumer — uses the function form with its original policy (network or 429, 3 retries, linear 1s backoff) against raw `AxiosError`s, since its finnhub client skips `simplifyError`. Its `vi.mock(.axios-retry.)` workaround is gone too. (`ts-retry-promise` in the WebSocket managers is a different domain — left as is.)

## Verified

- Typecheck: all 5 packages
- Tests: exchange 95, trading-signals 453, trading-strategies 259, messaging 84 — all green
- `tsc` emit + runtime smoke test of built `dist/` class
- Next.js docs build (compiles exchange from source via tsconfig paths)
- eslint clean; `tsx` demo scripts unaffected (esbuild 0.28 downlevels standard decorators)

### DIFF
--- a/packages/exchange/src/broker/trading212/api/Trading212API.ts
+++ b/packages/exchange/src/broker/trading212/api/Trading212API.ts
@@ -1,6 +1,6 @@
-import axios, {type AxiosDefaults, type AxiosError, type AxiosInstance} from 'axios';
-import axiosRetry from 'axios-retry';
+import axios, {type AxiosDefaults, type AxiosInstance} from 'axios';
 import {z} from 'zod';
+import {retry} from '../../../util/retry.js';
 import {simplifyError} from '../../../util/simplifyError.js';
 import {AccountCashSchema, AccountInfoSchema} from './schema/AccountSchema.js';
 import {HistoryOrderPageSchema, type HistoryOrder} from './schema/HistoryOrderSchema.js';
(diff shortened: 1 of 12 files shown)


packages/exchange/src/util/retry.ts
```diff
@@ -0,0 +1,49 @@
+import {SimplifiedHttpError} from './SimplifiedHttpError.js';
+
+type AsyncMethod<This, Args extends unknown[], Result> = (this: This, ...args: Args) => Promise<Result>;
+
+export interface RetryOptions {
```
bennycode, line 5: extract `RetryOptions` from axios-retry

vitest.config.base.ts
```diff
@@ -1,11 +1,31 @@
...
 export default defineConfig({
-  esbuild: {
-    // Allows using the "accessor" keyword in TypeScript:
-    // https://github.com/vitest-dev/vitest/issues/5976#issuecomment-2190804966
+  oxc: {
+    // Downlevels syntax Node cannot parse natively yet (e.g. the "accessor" keyword).
+    // Vite 8 (rolldown) transforms with oxc; the former `esbuild` options are ignored.
     target: 'es2022',
   },
+  plugins: [
```
bennycode, line 10: this is weird, find another alternative

packages/exchange/src/broker/trading212/api/Trading212API.ts
```diff
@@ -102,61 +63,71 @@ export class Trading212API {
...
 
   /** @see https://t212public-api-docs.redoc.ly/#operation/portfolio */
+  @retry({delayMs: 5_000})
   async getPositions() {
     const response = await this.#httpClient.get(URL.PORTFOLIO);
     return z.array(PositionSchema).parse(response.data);
   }
 
   /** @see https://t212public-api-docs.redoc.ly/#operation/positionByTicker */
+  @retry()
```
bennycode, line 87: make `delay` mandatory so we can see what's the retry delay directly at the function

PR #1204, opened 2026-07-20
chore(exchange): drop jsonabc and clean up dependency placement

bennycode:
## What

Dependency cleanup from a workspace-wide technology review:

- **`jsonabc` + `@types/jsonabc` removed** from exchange. The single call site was `CandleBatcher.toString()`, where `sortObj()` re-sorted an object literal that is already alphabetically sorted — and enforced to stay sorted by the `perfectionist/sort-objects` lint rule. Plain `JSON.stringify` produces byte-identical output; a comment now documents the invariant.
- **`ts-node` removed** from root devDependencies — nothing references it; `tsx` is the runner everywhere.
- **`@typedtrader/exchange` moved to devDependencies** in `@typedtrader/candles` — src only uses a type-only `Candle` import.

## Verification

- exchange: 89 tests passed (incl. CandleBatcher), dependency-cruiser clean
- candles: 52 tests passed

> Touches `package-lock.json` like #1202/#1203 — merge in any order, refresh with `npm install` on rebase.


packages/exchange/src/candle/CandleBatcher.ts
```diff
@@ -256,8 +255,9 @@ export class CandleBatcher extends EventEmitter<EventMap> {
   }
 
   static override toString(candle: BatchedCandle) {
+    // Deterministic output relies on the keys staying alphabetically sorted (enforced by perfectionist/sort-objects).
```
bennycode: remove comment

bennycode: Removed in bd8ba1c.

PR #1233, opened 2026-07-25
chore: migrate from ESLint and Prettier to Oxlint and oxfmt

bennycode:
Replaces ESLint with [Oxlint](https://oxc.rs/docs/guide/usage/linter) and Prettier with [oxfmt](https://oxc.rs/docs/guide/usage/formatter). Both are Rust binaries from the Oxc project.

| | before | after |
| --- | --- | --- |
| lint (488 files) | 9.1s | **1.3s** |
| format check (595 files) | n/a | **0.6s** |
| lint config files | 8 | **1** |

## Configuration

Eight ESLint configs (a shared `eslint.config.base.ts` plus one per package) collapse into a single root `oxlint.config.ts`, with the per-package ignore globs merged in.

Both configs are TypeScript, built with the `defineConfig` helper each package exports, so options are type checked and autocompleted, and the configs can carry comments explaining why a rule is set the way it is. Only `oxlint.config.ts` is auto-discovered by Oxlint (`.oxlintrc.ts` is not); oxfmt accepts either, so `oxfmt.config.ts` is used for consistency. Oxlint still marks JS/TS config files as experimental.

Prettier settings were carried over with `oxfmt --migrate=prettier`, which mapped every option except `requirePragma`, and that was set to its default of `false`.

The root `package.json` gains `"type": "module"`. Without it Node warns on every Oxlint and oxfmt run that the config parses as ESM but is not declared as such. The field only affects root-level files, and there are no root-level `.js` files.

`@oxlint/migrate` output was not used verbatim. It emitted ~200 lines, mostly dead `off` entries expanded from `eslint-config-prettier`, and listed plugins that are not installed, which made Oxlint fail to start. The config here is hand written.

## Rule coverage

35 of the 41 previous rules map to Oxlint natively. Six are dropped:

| rule | reason |
| --- | --- |
| `no-invalid-this` | superseded by `noImplicitThis`, already on via `strict: true` |
| `space-in-parens` | the formatter governs paren spacing |
| `strict` | requires `'use strict'`, meaningless in ESM |
| `one-var` | not implemented in Oxlint |
| `padding-line-between-statements` | not implemented in Oxlint, and the formatter does not insert blank lines after imports |
| `no-restricted-syntax` | the house rule forbidding explicit primitive and `Promise<primitive>` return types, dropped deliberately |

`padding-line-between-statements` and `one-var` are small stylistic losses with no replacement. Explicit primitive return types are no longer flagged.

## Type-aware linting

The five type-aware rules (`no-floating-promises`, `no-unnecessary-type-assertion`, `no-unsafe-argument`, `prefer-readonly`, `return-await`) run through `oxlint-tsgolint`, enabled by `options.typeAware` in the root config rather than a CLI flag. It bundles its own TypeScript, so the repo stays on 6.0.3.

Verified against a file with deliberate violations of each: Oxlint reports them at the same positions ESLint did.

## Scripts

Every package now has the same six scripts, and the root mirrors them:

```
fix        npm run fix:code && npm run fix:other
fix:code   oxlint . --fix || true
fix:other  oxfmt
lint       npm run lint:code && npm run lint:other
lint:code  oxlint .
lint:other oxfmt --list-different
```

The `|| true` is deliberate. `oxlint --fix` exits non-zero even for files it has fully cleaned, because when two rules want to rewrite the same span it applies one and still reports the other. Without it, `fix` would stop before formatting runs. The pre-commit hook has the same shape: it ignores the fix pass exit code and runs a separate verification pass to decide whether the commit proceeds.

## Incidental

`vitest.config.base.ts` had a comment violating `multiline-comment-style`. It went unnoticed because `eslint .` could not parse the file (`parserOptions.project: true` found no tsconfig for it), and the root `lint` script only linted two config files by name, so the repo root was effectively unlinted.

Three files pick up small formatting changes where oxfmt differs from Prettier: a wrapped union type in `utils/types.ts`, a long string moved to its own line in `vitest.config.ts`, and an object collapsed inside a README code block.

ESLint remains in `node_modules` as a peer dependency of `@stylistic/eslint-plugin` and `eslint-plugin-perfectionist`, which still run through Oxlint's JS plugin bridge. It is no longer a direct dependency and nothing invokes it.


packages/exchange/package.json
```diff
@@ -41,14 +41,13 @@
     "fix": "npm run fix:code && npm run fix:other",
     "fix:code": "oxlint . --fix || true",
     "fix:other": "oxfmt",
-    "lint": "npm run lint:types && npm run lint:code && npm run lint:other",
+    "lint": "npm run test:types && npm run lint:code && npm run lint:other",
```
bennycode: Remove `npm run test:types` from `lint` as it is part of `test` already.

bennycode: Done in 85213fd. Removed from all six packages and from the root `lint`, which no longer needs a lerna call at all.

Verified the counts: `npm run lint` is now 1 oxlint + 1 oxfmt + 0 tsc, and `npm test` still runs `tsc --noEmit` for all six packages, so type coverage is unchanged.

lefthook.yml
```diff
@@ -3,10 +3,18 @@ pre-commit:
...
     lint:
       glob: 'packages/**/*.{js,jsx,ts,tsx}'
-      run: npx eslint {staged_files} --fix
+      # Rule fixes first, formatting last, then a verification pass that decides whether the
+      # commit proceeds. The fix run's own exit code is deliberately ignored: when two rules
+      # want to rewrite the same span, oxlint applies one and still reports the other, so it
+      # exits non-zero even for files it has fully cleaned. Only the final run is meaningful.
+      # This command and `format` have disjoint globs, so they never race despite `parallel`.
+      run: |
+        npx oxlint --fix {staged_files} >/dev/null 2>&1 || true
```
bennycode: Can't you just do `npx oxlint --fix {staged_files} || true`?

oxfmt.config.ts
```diff
@@ -0,0 +1,23 @@
+import {defineConfig} from 'oxfmt';
+
+/**
+ * Formatting for every file type in the monorepo, code included. Oxfmt discovers this file from
```
bennycode: remove comment

oxfmt.config.ts
```diff
@@ -0,0 +1,23 @@
...
+export default defineConfig({
+  arrowParens: 'avoid',
+  bracketSameLine: true,
+  bracketSpacing: false,
+  printWidth: 120,
+  proseWrap: 'never',
+  semi: true,
+  singleQuote: true,
+  /*
+   * Several packages group their scripts by purpose rather than alphabetically. Sorting would
```
bennycode: i want sorted packageJSON

oxlint.config.ts
```diff
@@ -0,0 +1,99 @@
+import {defineConfig} from 'oxlint';
+
+/**
+ * Single lint config for the whole monorepo. Oxlint resolves it from any package directory, so
```
bennycode: remove comment

oxlint.config.ts
```diff
@@ -0,0 +1,99 @@
...
+ * per-package scripts are a bare `oxlint .`.
+ */
+export default defineConfig({
+  /*
+   * Oxlint enables its `correctness` rules by default. Switching the category off keeps this
```
bennycode: remove this comment

lefthook.yml
bennycode: Yes, that works, and it is what the hook does now.

The redirection was only suppressing duplicate output: the fix pass and the verification pass both report anything unfixable, so a blocked commit now prints each error twice. The hook still blocks correctly. If the double print gets annoying, putting `>/dev/null 2>&1` back on the fix line is the only thing that changes.

oxfmt.config.ts
bennycode: Removed.

oxfmt.config.ts
bennycode: Done, `sortPackageJson` is now `true`.

My justification for setting it to `false` was wrong, so worth correcting: it only reorders top-level keys into npm's canonical order and leaves the contents of `scripts` untouched. I checked `exchange` and `trading-strategies`, which group their scripts by purpose rather than alphabetically, and the order is byte-identical before and after.

oxlint.config.ts
bennycode: Removed.

oxlint.config.ts
bennycode: Removed.

I kept the remaining three comments in this file: the `jsPlugins` and `typeAware` lines, and the `no-redeclare` and `perfectionist` rationales carried over verbatim from `eslint.config.base.ts`. Say the word if you want those gone too.

PR #1236, opened 2026-07-27
feat(trading-signals): add AtrTrail indicator (ATR_TRAIL)

bennycode:
The feature at the heart of #1131 — a trailing stop whose width is sized from the instrument's own volatility — extracted as a standalone, publicly exported indicator in the `trading-signals` library.

## What it does
- Trail width = `multiplier × ATR%` (via `NATR`), so a volatile name automatically gets room to breathe while a calm one gets a tight stop — no hand-tuned percentage.
- The peak **ratchets upward only** and the emitted stop **never decreases** — unlike `ChandelierExit`, whose highest high rolls out of a fixed window and whose stop can drop again.
- Two modes:
  - **`FROZEN`** (default): the width is measured once when the ATR warms up and stays fixed — a plain percentage trail sized sensibly from recent history.
  - **`ROLLING`**: the width keeps adapting to the live ATR, but a volatility spike only widens the trail for future peaks — it never loosens a stop already in place.

## Usage
```ts
const trail = new AtrTrail({interval: 14, multiplier: 2, mode: AtrTrailMode.ROLLING});

const result = trail.add({high: 200, low: 190, close: 195});
// => {
//   peak: 200,      // ratcheting highest high
//   stop: 160,      // never-decreasing stop
//   trail: 160,     // raw candidate: peak * (1 - trailPct/100)
//   trailPct: 20,   // width from multiplier x ATR%
// }
// null during warm-up
```

`trail` and `stop` diverge in `ROLLING` mode when a volatility spike pushes the raw candidate below the committed stop — both are exposed so consumers see the ratchet at work.

## Notes
- Extends `TechnicalIndicator<AtrTrailResult, HighLowClose<number>>` (multi-value result, like `ChandelierExit`/`SuperTrend`), correct bidirectional `replace()` via a one-deep state snapshot, `getRequiredInputs()` delegated to the internal `NATR`.
- 7 unit tests: exact-value sequences for both modes (including the `trail`/`stop` divergence), volatile-vs-calm width comparison, defaults, `NotEnoughDataError`, and bidirectional replace per mode. 100% coverage maintained.
- #1131 (`AtrTrailStrategy`) can adopt this indicator later instead of carrying the trail logic inline.

### DIFF
--- a/packages/trading-signals/src/trend/ATR_TRAIL/AtrTrail.ts
+++ b/packages/trading-signals/src/trend/ATR_TRAIL/AtrTrail.ts
@@ -1,0 +1,89 @@
+import type {HighLowClose} from '../../base/Candle.type.js';
+import {IndicatorSeries} from '../../base/Indicator.js';
+import {NATR} from '../../volatility/NATR/NATR.js';
+
+export const AtrTrailMode = {
+  /** Size the trail width once when the ATR warms up, then leave it alone. */
+  FROZEN: 'FROZEN',
+  /** Re-size the trail width from the live ATR on every candle. */
+  ROLLING: 'ROLLING',
+} as const;
+
+export type AtrTrailModes = (typeof AtrTrailMode)[keyof typeof AtrTrailMode];
+
+export type AtrTrailConfig = {
+  /** ATR lookback used to measure the instrument's volatility (default: 14) */
+  interval?: number;
+  /** Sizing behavior of the trail width (default: FROZEN) */
+  mode?: AtrTrailModes;
+  /** How many ATR% the trail sits below the peak (default: 3, Chandelier convention) */
+  multiplier?: number;
+};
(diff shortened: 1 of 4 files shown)


packages/trading-signals/src/trend/ATR_TRAIL/AtrTrail.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.mode = mode;
+    this.multiplier = multiplier;
+    this.#natr = new NATR(interval);
+  }
+
+  override getRequiredInputs() {
+    return this.#natr.getRequiredInputs();
+  }
+
+  update(candle: HighLowClose<number>, replace: boolean) {
```
bennycode, line 79: @claude will it only return the stop price and never the trailing up price?

PR #1242, opened 2026-07-31
fix(trading-signals): use correct comparand when replacing ROC values

Sahilhawal:
## Problem

`ROC` recomputes its comparand (the price `interval` bars ago) from the current window head on every `update()`. Once more than `interval` prices have been added, the oldest price has already been evicted from the rolling window, so on a `replace()` `prices[0]` points one bar too recent. Replacing the latest bar — even with the same value — therefore changes the result.

### Reproduction

```ts
const roc = new ROC(3);
[10, 11, 12, 13].forEach(p => roc.add(p));
roc.getResultOrThrow(); // 0.3        (13 - 10) / 10  ✅
roc.replace(13);        // same value
roc.getResultOrThrow(); // 0.18181…   ❌ expected 0.3
```

The existing `replace()` test only replaces the first stable bar (before any eviction), where `prices[0]` still happens to be the correct comparand — which is why this went unnoticed.

## Fix

Persist the comparand captured when a bar is first added (`#comparePrice`) and reuse it on `replace()` instead of re-reading the window head.

## Tests

- Added a regression test that replaces a bar after the window has advanced and asserts the result matches the equivalent non-replaced series.
- Full `trading-signals` suite passes (541 tests) and `tsc --noEmit` is clean.

### DIFF
--- a/packages/trading-signals/src/momentum/ROC/ROC.test.ts
+++ b/packages/trading-signals/src/momentum/ROC/ROC.test.ts
@@ -143,5 +143,19 @@ describe('ROC', () => {
 
       expect(roc.getResultOrThrow().toFixed(2)).toEqual(expectation.toFixed(2));
     });
+
+    it('keeps the correct comparand when replacing after the window has advanced', () => {
+      const interval = 3;
+
+      const withReplace = new ROC(interval);
+      [10, 11, 12, 13].forEach(price => withReplace.add(price));
+      withReplace.replace(13);
+
+      const withoutReplace = new ROC(interval);
+      [10, 11, 12, 13].forEach(price => withoutReplace.add(price));
+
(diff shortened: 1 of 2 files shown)


Sahilhawal:
@bennycode could you please take a look at this PR
thanks!

bennycode:
Hi @Sahilhawal, thank you for your contribution! I took a look at your changes, and they make perfect sense. I'm really glad you fixed the bug and submitted a PR for it. I'll merge your PR and include it in the next release. 👍

bennycode:
@Sahilhawal can you test the fix with trading-signals (8.1.0)?

Sahilhawal:
yep sure @bennycode  
let me check

Sahilhawal:
@bennycode tested on 8.1.0, works

PR #1243, opened 2026-07-31
fix(trading-signals): recalculate Bollinger Bands on replace()

bennycode:
## Problem

`BollingerBands.update()` gated its calculation on `dropOut` — the price evicted by `pushUpdate()`:

```ts
const dropOut = pushUpdate(this.prices, replace, price, this.interval);
if (dropOut) {
  // calculate bands
}
```

Only an `add()` evicts a price. A `replace()` swaps the last element without changing the array length, so nothing is evicted, `dropOut` is `null`, and the method returns `null` — after the `if (replace)` block has already rewound `this.result`. Replacing the latest price of a warmed-up indicator therefore **erases the result**:

```ts
const bb = new BollingerBands(5, 2);
[10, 11, 12, 13, 14, 15].forEach(p => bb.add(p));
bb.getResult()?.middle; // 13
bb.replace(999);
bb.getResult()?.middle; // undefined  ❌
```

The same gate checks the evicted price for *truthiness*, so a dropped-out price of `0` also suppresses the calculation:

```ts
const bb = new BollingerBands(3, 2);
[0, 1, 2, 3].forEach(p => bb.add(p));
bb.getResult(); // null  ❌ expected middle 2
```

`BollingerBandsWidth` wraps `BollingerBands` and inherited both bugs.

## Fix

Latch a private `#hasEvictedPrice` flag on the first eviction and gate the calculation on that. The flag survives a `replace()`, so a replacement recalculates over the current window, and the latch is set by an explicit `!== null` check rather than truthiness.

## Scope

Warm-up timing is deliberately unchanged here. `BollingerBands` still emits its first band one bar later than its own reference fixture expects (`middle[19]` in `src/fixtures/BB/data.json` is `74.3 = avg(prices[0..19])`, but the indicator is not stable until the 21st price). That is a separate behavior change and gets its own PR.

## Tests

Two regression tests, both verified to fail on `main` and pass here:
- a bidirectional replace test — replace changes the bands, replacing back restores them
- a zero drop-out test

Full suite passes (543 tests), coverage stays at 100%, `npm run lint` and `tsc --noEmit` clean.

### DIFF
--- a/packages/trading-signals/src/volatility/BBANDS/BollingerBands.test.ts
+++ b/packages/trading-signals/src/volatility/BBANDS/BollingerBands.test.ts
@@ -161,6 +161,20 @@ describe('BollingerBands', () => {
       expect(restoredResult?.middle, 'replacing back restores the original bands').toBe(originalResult.middle);
     });
 
+    it('changes nothing when the latest price is replaced with the same value', () => {
+      const bb = new BollingerBands(5, 2);
+      const prices = [81.59, 81.06, 82.87, 83.0, 83.61, 83.15] as const;
+
+      prices.forEach(price => bb.add(price));
+
+      const resultBefore = bb.getResultOrThrow();
+
+      bb.replace(83.15);
+
+      expect(bb.getResultOrThrow(), 'replacing a price with itself is a no-op').toEqual(resultBefore);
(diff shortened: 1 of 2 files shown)


bennycode:

PR #1244, opened 2026-07-31
fix(trading-signals): emit the first Bollinger Band once the interval is filled

bennycode:
> Stacked on #1243 — review that one first. Base will retarget to `main` once it merges.

## Problem

`BollingerBands` only calculated after a price had been **evicted** from the window, so it needed `interval + 1` inputs and never emitted the first calculable band.

Its own reference fixture proves it. In `src/fixtures/BB/data.json` with `interval = 20`:

```
middle[19] = 74.3 = avg(prices[0..19])
```

The reference expects a band at the 20th price, but the indicator was not stable until the 21st. Because the test loops with `if (!bb.isStable) return;`, that expected value was **silently skipped**, and everything after it lined up one bar late by coincidence. The Tulip Indicators test skipped its `index 4` for the same reason.

```ts
const bb = new BollingerBands(5);
const sma = new SMA(5);
[10, 11, 12, 13, 14].forEach(p => { bb.add(p); sma.add(p); });

sma.isStable; // true
bb.isStable;  // false  ❌ same window, one bar later
```

## Fix

Gate on the window being full (`this.prices.length === this.interval`) instead of on the eviction. This also removes the `#hasEvictedPrice` latch introduced in #1243, which existed only to keep that PR free of any timing change.

## Behavior change

Consumers now receive their first `BollingerBands` and `BollingerBandsWidth` result one bar earlier. Values at every subsequent bar are unchanged. `getRequiredInputs()` already returned `interval`, so it is now accurate rather than off by one.

## Tests

- Both reference tests (fixture and Tulip) now assert the value they previously skipped, and pass unchanged otherwise.
- `BollingerBandsWidth` gains the leading `'0.19'` its series was missing, and switches to the repo's usual `getRequiredInputs() - 1` offset.
- New warm-up test pinning that the fifth price of a `BollingerBands(5)` produces the first band.

`trading-strategies` (259 tests) passes unchanged. Full `trading-signals` suite passes (544 tests), coverage stays at 100%, lint and `tsc --noEmit` clean.

### DIFF
--- a/packages/trading-signals/src/volatility/BBANDS/BollingerBands.test.ts
+++ b/packages/trading-signals/src/volatility/BBANDS/BollingerBands.test.ts
@@ -49,6 +49,20 @@ describe('BollingerBands', () => {
       expect(bb.deviationMultiplier).toBe(2);
     });
 
+    it('emits the first result as soon as the interval is filled', () => {
+      const bb = new BollingerBands(5, 2);
+      const prices = [10, 11, 12, 13, 14] as const;
+
+      prices.slice(0, -1).forEach(price => bb.add(price));
+
+      expect(bb.isStable, 'four prices do not fill an interval of five').toBe(false);
+
+      const result = bb.add(prices[4]);
+
+      expect(bb.isStable, 'the fifth price completes the interval').toBe(true);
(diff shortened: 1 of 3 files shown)


bennycode:

PR #1256, opened 2026-08-01
fix(trading-signals): roll back TDS setup state on replace()

bennycode:
## Problem

TDS tracks a running `setupCount` and `setupDirection`, advancing them on every update and resetting them once a 9-bar setup completes. `replace()` popped the last close but left those mutations in place, so the replaced bar was counted a second time.

**The count runs away:**

```ts
const tds = new TDS();
[10, 10, 10, 10].forEach(c => tds.add(c));
tds.add(15);
tds.add(16);
tds['setupCount']; // 2

tds.replace(17);
tds['setupCount']; // 3  ❌ still two bars, counted three times
```

Because a replacement can push the count past the 9-bar threshold, it could fabricate a completed setup on an indicator that was not stable before the call.

**A broken setup stays emitted:**

```ts
// ... 9 consecutive rising closes complete a bullish setup
tds.getResult(); // 1

tds.replace(5);  // the last bar is now a big down bar
tds.getResult(); // 1  ❌ the setup no longer completes
```

## Fix

Snapshot `setupCount`/`setupDirection` before each `add()` and restore them when replacing, so the replaced bar re-runs from the state it originally found.

A replacement that no longer completes a setup also has to withdraw the value it previously emitted. That is what `rollbackLastResult()` on `IndicatorSeries` is for — its doc comment names exactly this case ("sparse indicators whose `replace()` can invalidate a prior emission without producing a new one").

## Tests

Four new tests. Three fail on `main` and pass here:
- setup count not advanced twice (`3` vs `2`)
- a replaced series matches the equivalent add-only series
- a completed setup is withdrawn when the replacement breaks it

The fourth pins the direction-reversal path, which the rollback could plausibly regress.

Full suite passes (545 tests), coverage stays at 100%, lint and `tsc --noEmit` clean.

### DIFF
--- a/packages/trading-signals/src/momentum/TDS/TDS.test.ts
+++ b/packages/trading-signals/src/momentum/TDS/TDS.test.ts
@@ -111,6 +111,23 @@ describe('TDS', () => {
       expect(result).toBeNull();
     });
 
+    it('changes nothing when the latest close is replaced with the same value', () => {
+      const tds = new TDS();
+      const closes = [10, 10, 10, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19] as const;
+
+      closes.forEach(close => tds.add(close));
+
+      const resultBefore = tds.getResult();
+      const countBefore = tds['setupCount'];
+      const directionBefore = tds['setupDirection'];
+
+      tds.replace(19);
+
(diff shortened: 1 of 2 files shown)


bennycode:

PR #1257, opened 2026-08-01
fix(trading-signals): restore PSAR state on replace()

bennycode:
## Problem

PSAR carries six mutable fields from bar to bar — `acceleration`, `extreme`, `lastSar`, `isLong`, `previousCandle` and `prePreviousCandle`. `replace()` overwrote none of them, so every value the discarded candle produced stayed in the indicator and kept influencing later bars.

The discarded candle stays visible in the result:

```ts
const psar = new PSAR({accelerationMax: 0.2, accelerationStep: 0.02});
candles.slice(0, -1).forEach(c => psar.add(c));
psar.add({high: 999, low: 1});          // decoy
psar.replace(candles.at(-1));           // replaced with the real candle
// result: 999  ❌ expected 85.07 — the value an add-only series produces
```

Even replacing a candle with **itself** shifted the result, because the second-candle branch re-ran against already-advanced state:

```ts
psar.add(candles[0]);
psar.add(candles[1]);                   // 82.15
psar.replace(candles[1]);               // 82.1198  ❌ same input, different output
```

## Fix

Snapshot the six fields before each `add()` and restore them when replacing, so the replacement re-runs from the state the original candle saw. Those six are the complete set `update()` reads before deciding the next SAR, so the snapshot fully determines the transition.

## Removed early return

```ts
const notEnoughData = !this.previousCandle || this.lastSar === null;
if (replace && notEnoughData) {
  this.previousCandle = candle;
  return null;
}
```

This existed to stop `replace()` reading corrupted state, and it had a side effect: replacing the **second** candle returned `null` instead of the first SAR, because `lastSar` was still `null` at that point in the original bar. With the state restored, the ordinary first-candle branch covers the genuine not-enough-data case, so the special case is gone. Both existing tests that pin this behavior (`handles replace flag for second candle`, `handles replace flag with notEnoughData`) still pass unchanged.

## Tests

Three tests, all verified to fail on `main`:
- replacing a candle with itself is now exact (tightened from a 1% tolerance to `toBe`)
- a replaced series reproduces the equivalent add-only series
- the same, with a decoy that forces a long-to-short **reversal**, which resets acceleration and the extreme point

Full suite passes (543 tests), coverage stays at 100%, lint and `tsc --noEmit` clean.

### DIFF
--- a/packages/trading-signals/src/trend/PSAR/PSAR.test.ts
+++ b/packages/trading-signals/src/trend/PSAR/PSAR.test.ts
@@ -48,17 +48,23 @@ describe('PSAR', () => {
     });
   });
 
-  it('handles replace correctly', () => {
-    const psar = new PSAR({accelerationMax: 0.2, accelerationStep: 0.02});
+  it('changes nothing when the latest candle is replaced with the same values', () => {
+    const candles = testData.map(({high, low}) => ({high, low}));
 
-    // Add first two candles
-    psar.add({high: testData[0].high, low: testData[0].low});
-    const initialResult = psar.add({high: testData[1].high, low: testData[1].low});
+    // Swept over every length because the carried state only diverges at some of them.
+    for (let length = 1; length <= candles.length; length++) {
+      const psar = new PSAR({accelerationMax: 0.2, accelerationStep: 0.02});
 
-    // Replace the second candle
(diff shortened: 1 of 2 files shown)


bennycode:

PR #1258, opened 2026-08-01
fix(trading-signals): restore ZigZag swing state on replace()

bennycode:
## Problem

ZigZag carries `isUp`, `highestExtreme` and `lowestExtreme` across candles and emits a pivot only when the trend reverses. `replace()` left all three holding whatever the replaced candle produced, so a discarded candle kept setting the extremes:

```ts
const zigzag = new ZigZag({deviation: 15});
candles.slice(0, -1).forEach(c => zigzag.add(c));
zigzag.add({high: 500, low: 400});     // decoy
zigzag.replace(candles.at(-1));        // replaced with the real candle
zigzag.getResult(); // 500  ❌ expected 40 — what an add-only series produces
```

A pivot also survived being invalidated. ZigZag only emits on a reversal, so when a replacement means no reversal happened, the previously reported swing point has to be withdrawn:

```ts
zigzag.add({high: 88, low: 10});   // deep low ends the uptrend → emits 88
zigzag.replace({high: 88, low: 85}); // quiet candle, no reversal after all
zigzag.getResult(); // 10  ❌ the reversal was undone but the pivot lingered
```

## Fix

Snapshot the three fields before each `add()` and restore them when replacing. Withdraw the last emission with `rollbackLastResult()`, whose doc comment names this exact case ("sparse indicators whose `replace()` can invalidate a prior emission without producing a new one").

## Tests

ZigZag had **no** `replace()` test before this. Three added:
- a replaced series reproduces the equivalent add-only series (fails on `main`: `500` vs `40`)
- an invalidated pivot is withdrawn (fails on `main`: `10` vs `0`)
- replacing before any candle behaves like adding the first candle

Full suite passes (544 tests), coverage stays at 100%, lint and `tsc --noEmit` clean.

### DIFF
--- a/packages/trading-signals/src/trend/ZIGZAG/ZigZag.test.ts
+++ b/packages/trading-signals/src/trend/ZIGZAG/ZigZag.test.ts
@@ -43,6 +43,74 @@ describe('ZigZag', () => {
     });
   });
 
+  describe('replace', () => {
+    const highs = [
+      -8, -4, -1, 9, 8, 7, 6, 5, 4, 3, 2, 1, 11, 22, 33, 44, 55, 66, 77, 88, 88, 71, 61, 51, 41, 51, 61, 71, 81, 91,
+    ] as const;
+
+    const lows = [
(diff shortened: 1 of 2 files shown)


bennycode:

PR #1259, opened 2026-08-01
fix(trading-signals): report the correct number of required inputs for ROC

bennycode:
## Problem

ROC compares the current price against the price `interval` bars back, so it needs `interval + 1` inputs before it can produce anything. `getRequiredInputs()` returned just the interval:

```ts
const roc = new ROC(5);
roc.getRequiredInputs(); // 5

[1, 2, 3, 4, 5].forEach(p => roc.add(p));
roc.isStable;            // false  ❌ it said 5 was enough
```

`NotEnoughDataError` is constructed from the same value, so it also quoted the wrong minimum (`A minimum of "5" inputs is required` when six are needed).

## Precedent

`MOM` and `VROC` do the same "value n periods ago" arithmetic and both already size their history to `interval + 1` and report it:

```ts
// VROC
this.#historyLength = interval + 1;
override getRequiredInputs() {
  return this.#historyLength;
}
```

ROC was the odd one out of the three.

## Scope

Only the reported number was wrong — ROC's actual behavior (first result on the `interval + 1`th input) is correct and unchanged. No values move.

## Tests

- The Tulip test's offset moves to the repo's usual `getRequiredInputs() - 1`, so it asserts the same expectations against the same bars as before. With the old return value it now fails (`'0.02'` vs `'0.01'`), so it pins the number rather than absorbing it.
- New test asserting the indicator becomes stable after exactly the number of inputs it reports, which fails on `main`.

Full suite passes (542 tests), `trading-strategies` passes unchanged (259 tests), coverage stays at 100%, lint and `tsc --noEmit` clean.

## Still open

Separate from this: replacing the 5th bar of a `ROC(5)` still flips it from unstable to stable with a fabricated result. That needs the window-sizing change (holding `interval + 1` prices like MOM/VROC), not a warm-up count fix.

### DIFF
--- a/packages/trading-signals/src/momentum/ROC/ROC.test.ts
+++ b/packages/trading-signals/src/momentum/ROC/ROC.test.ts
@@ -20,7 +20,7 @@ describe('ROC', () => {
 
       const interval = 5;
       const roc = new ROC(interval);
-      const offset = roc.getRequiredInputs();
+      const offset = roc.getRequiredInputs() - 1;
 
       prices.forEach((price, i) => {
         roc.add(price);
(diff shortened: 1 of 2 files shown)


bennycode:

PR #1260, opened 2026-08-01
fix(trading-signals): hold the ROC comparand inside the price window

bennycode:
> Stacked on #1259 — review that one first. Base will retarget to `main` once it merges.

## Problem

ROC kept only `interval` prices, so the bar it compares against had already been evicted from the window by the time it was needed. #1242 worked around that by caching it in `#comparePrice`, refreshed only on an `add()`.

That left one boundary case wrong. Replacing the `interval`-th bar re-reads a comparand that does not exist yet, fabricating a result and turning an unstable indicator stable:

```ts
const roc = new ROC(3);
[10, 11, 12].forEach(p => roc.add(p));
roc.isStable;      // false — correct, ROC(3) needs 4 prices
roc.replace(12.5);
roc.isStable;      // true  ❌ result 0.25 out of thin air
```

A `replace()` does not add a bar, so it can never move an indicator from unstable to stable.

## Fix

Size the window to `interval + 1` and read `prices[0]`, the way `MOM` and `VROC` already do for the same arithmetic. The window becomes the only state ROC carries, so `pushUpdate()` makes `replace()` correct on its own and the cache disappears.

This is what `packages/trading-signals/CLAUDE.md` prescribes: *"if the result is a pure function of the candle window, `pushUpdate()` handles it"*. ROC's result is a pure function of the last `interval + 1` prices.

## Verification

**Values are unchanged.** Checked every emission against the textbook definition `(p[k] - p[k-interval]) / p[k-interval]` — 2322 emitted values across intervals 1 to 12, zero mismatches.

**`replace()` now always matches an add-only series.** Across 84 scenarios (intervals 1–6 × every series length), comparing `add(decoy) + replace(real)` against adding the real series:

| | mismatches |
|---|---|
| before (cached comparand) | **6 of 84** — one per interval, each at the boundary |
| after | **0 of 84** |

## Rewritten test

`returns a valid result when replacing values` fed 5 prices into a `ROC(5)` and asserted a result — it was pinning this very bug as expected behaviour. It only passed because `toFixed(2)` rounded the actual `0.0153` and the expected `0.01912` to the same `"0.02"`. It is now a bidirectional replace test on a warmed-up indicator, asserting exact values.

Added a test for the boundary case itself, which fails on `main`.

Full suite passes (543 tests), `trading-strategies` passes unchanged (259 tests), coverage stays at 100%, lint and `tsc --noEmit` clean.

### DIFF
--- a/packages/trading-signals/src/momentum/ROC/ROC.test.ts
+++ b/packages/trading-signals/src/momentum/ROC/ROC.test.ts
@@ -144,18 +144,30 @@ describe('ROC', () => {
 
   describe('update', () => {
     it('returns a valid result when replacing values', () => {
-      const prices = [81.59, 81.06, 82.87, 83.0, 83.61] as const;
-      const updatePrice = 82.84;
-      const expectation = 0.01911999019;
+      const prices = [81.59, 81.06, 82.87, 83.0, 83.61, 83.15] as const;
 
-      const interval = 5;
-      const roc = new ROC(interval);
+      const roc = new ROC(5);
 
       prices.forEach(price => roc.add(price));
 
-      roc.replace(updatePrice);
+      const originalResult = roc.getResultOrThrow();
+      const replacedResult = roc.replace(82.84);
 
-      expect(roc.getResultOrThrow().toFixed(2)).toEqual(expectation.toFixed(2));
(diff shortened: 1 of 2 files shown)


bennycode:

PR #1261, opened 2026-08-01
fix(trading-signals): correct replace() across six indicators

bennycode:
Supersedes #1243, #1244, #1245, #1256, #1257, #1258, #1259, #1260 — same content, one PR. Eight commits, one per fix, each green on its own.

## How these were found

A differential harness over all 62 exported indicators, testing two properties:

1. a series with a decoy final bar plus `replace()` must equal the equivalent add-only series
2. `replace()` must never flip `isStable` from false to true, since it adds no input

That flagged 5 drift failures and 2 stability failures. A later sweep added a third property — replacing the last input with an **identical** value must change nothing — which failed on exactly the same six indicators, at 6 of 2480 scenarios.

## What was wrong

| Indicator | Defect |
|---|---|
| **BollingerBands** | `replace()` returned `null` and wiped the result to `undefined`; a dropped-out price of `0` suppressed the calculation; first band emitted one bar late |
| **BollingerBandsWidth** | inherited all three |
| **PVT** | replaced candle counted twice (`372.7` vs `281.8`) |
| **TDS** | setup counters never rolled back — a replacement could fabricate a completed setup |
| **PSAR** | six mutable fields never rolled back — a discarded candle's high of `999` stayed in the result |
| **ZigZag** | swing state never rolled back; had **no** `replace()` test at all |
| **ROC** | `getRequiredInputs()` off by one; replacing at the warm-up boundary fabricated a result |

Two shapes of root cause. Window-based indicators (BollingerBands, ROC) gated on the wrong thing or evicted the value they still needed — fixed by making the window the only state, so `pushUpdate()` handles `replace()` on its own, as `packages/trading-signals/CLAUDE.md` prescribes. State-machine indicators (PVT, TDS, PSAR, ZigZag) carried mutations across a replacement — fixed by snapshotting and restoring.

## Behaviour change

`BollingerBands` and `BollingerBandsWidth` now emit their first result one bar earlier. Values at every later bar are unchanged. This is a fix, not a tweak: `src/fixtures/BB/data.json` has `middle[19] = 74.3 = avg(prices[0..19])`, but the indicator stayed unstable until the 21st price. The test looped with `if (!bb.isStable) return;`, so that expected value was silently skipped. The Tulip test skipped its index 4 for the same reason. Both now assert the value they were skipping.

`ROC.getRequiredInputs()` returns `interval + 1`, matching `MOM` and `VROC`. Emitted values are unchanged.

## Verification

- **No Tulip divergence.** Every reference array is byte-identical to `main`. BollingerBands now asserts 11 bars against Tulip instead of 10, all matching — it moved toward the reference. ROC's mapping is unchanged: same bars, same expectations. PSAR's Tulip test only uses `add()`.
- **ROC values checked against the definition** `(p[k] - p[k-interval]) / p[k-interval]` — 2322 emissions across intervals 1–12, zero mismatches.
- **BBW tightened from 2 to 8 decimals.** TradingView's 2dp cannot separate neighbouring readings (the first two bars both display `0.19` while actually being `0.1853` and `0.1945`). The tighter values are derived from the definition rather than read back from our output, and each still rounds to the TradingView figure.
- **All three harness properties clean** across 62 indicators after the fixes.
- 563 tests, coverage stays at 100%, lint and `tsc --noEmit` clean, `trading-strategies` unchanged at 259 tests.

## Tests

Every new test was verified to fail without its fix. All six indicators now have an explicitly named same-value test; where the state only diverges at certain series lengths (ROC, PSAR, ZigZag) the test sweeps every length, because a single fixed length silently passed.

### DIFF
--- a/packages/trading-signals/src/base/Indicator.ts
+++ b/packages/trading-signals/src/base/Indicator.ts
@@ -25,10 +25,40 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
 /**
  * Implements common update behaviour among indicators.
  */
-export abstract class TechnicalIndicator<Result, Input> implements Indicator<Result, Input> {
+export abstract class TechnicalIndicator<
+  Result,
+  Input,
+  State extends object = Record<string, never>,
+> implements Indicator<Result, Input> {
   protected result: Result | undefined;
+  /**
+   * Single container for every mutable field (beyond the result) that `update()` may change.
+   * Keeping such fields here instead of in loose class fields lets the base class rewind them
+   * generically on `replace()` (see {@link trackState}) and makes two indicator instances
+   * comparable via {@link getState}.
+   */
+  protected state: State = {} as State;
+  #previousState?: State;
 
   abstract getRequiredInputs(): number;
+
+  protected trackState(replace: boolean) {
+    if (replace) {
+      if (this.#previousState !== undefined) {
+        this.state = structuredClone(this.#previousState);
+      }
+    } else {
+      this.#previousState = structuredClone(this.state);
+    }
+  }
+
+  /**
+   * Snapshot of the result and the mutable state, so two indicator instances fed the same
+   * inputs can be compared.
+   */
+  getState() {
+    return structuredClone({...this.state, result: this.result});
+  }
 
   getResult() {
     try {
@@ -68,7 +98,10 @@ export abstract class TechnicalIndicator<Result, Input> implements Indicator<Res
 /**
  * Tracks results of an indicator over time.
  */
-export abstract class IndicatorSeries<Input = number> extends TechnicalIndicator<number, Input> {
+export abstract class IndicatorSeries<
+  Input = number,
+  State extends object = Record<string, never>,
+> extends TechnicalIndicator<number, Input, State> {
   protected previousResult?: number;
 
   protected setResult(value: number, replace: boolean) {
@@ -101,7 +134,8 @@ export abstract class IndicatorSeries<Input = number> extends TechnicalIndicator
 export abstract class TrendIndicatorSeries<
   Input = number,
   SignalState = TradingSignals,
-> extends IndicatorSeries<Input> {
+  State extends object = Record<string, never>,
+> extends IndicatorSeries<Input, State> {
   protected abstract calculateSignalState(result?: number | null | undefined): SignalState;
   #previousSignalState?: SignalState;
 
--- a/packages/trading-signals/src/base/Indicator.ts
+++ b/packages/trading-signals/src/base/Indicator.ts
@@ -44,7 +44,13 @@ export abstract class TechnicalIndicator<
 
   protected trackState(replace: boolean) {
     if (replace) {
-      if (this.#previousState !== undefined) {
+      if (this.#previousState === undefined) {
+        /*
+         * A replacement before any input still needs a rollback baseline, so repeated
+         * replacements of the first logical input remain replacements.
+         */
+        this.#previousState = structuredClone(this.state);
+      } else {
         this.state = structuredClone(this.#previousState);
       }
     } else {
--- a/packages/trading-signals/src/base/Indicator.ts
+++ b/packages/trading-signals/src/base/Indicator.ts
@@ -31,12 +31,6 @@ export abstract class TechnicalIndicator<
   State extends object = Record<string, never>,
 > implements Indicator<Result, Input> {
   protected result: Result | undefined;
-  /**
-   * Single container for every mutable field (beyond the result) that `update()` may change.
-   * Keeping such fields here instead of in loose class fields lets the base class rewind them
-   * generically on `replace()` (see {@link trackState}) and makes two indicator instances
-   * comparable via {@link getState}.
-   */
   protected state: State = {} as State;
   #previousState?: State;
 
--- a/packages/trading-signals/src/fixtures/testIndicatorContract.ts
+++ b/packages/trading-signals/src/fixtures/testIndicatorContract.ts
@@ -1,17 +1,10 @@
 import {describe, expect, it} from 'vitest';
+import type {TechnicalIndicator} from '../base/Indicator.js';
 import {NotEnoughDataError} from '../error/NotEnoughDataError.js';
 
 type IndicatorContractOptions<Input> = {
   /** Creates a fresh instance of the indicator under test */
-  create: () => {
-    add(input: Input): unknown;
-    getRequiredInputs(): number;
-    getResult(): unknown;
-    getResultOrThrow(): unknown;
-    getState(): object;
-    isStable: boolean;
-    replace(input: Input): unknown;
-  };
+  create: () => TechnicalIndicator<unknown, Input, object>;
   /** Input that must lead the indicator into a different state than the last of `inputs` */
   divergentInput: Input;
   /** Input series fed before exercising replace(); the last input is the one being replaced */
@@ -46,6 +39,18 @@ export function testIndicatorContract<Input>({create, divergentInput, inputs}: I
         indicator.add(input);
         expectNoResult(index + 1);
       });
+    });
+
+    it('yields a result once the warm-up is complete', () => {
+      const indicator = create();
+
+      for (const input of inputs) {
+        indicator.add(input);
+      }
+
+      expect(indicator.isStable, 'the contract inputs leave the indicator stable').toBe(true);
+      expect(indicator.getResult(), 'a warmed-up indicator never yields undefined').not.toBeUndefined();
+      expect(indicator.getResultOrThrow(), 'a stable indicator yields an actual result').not.toBeNull();
     });
   });
 
--- a/packages/trading-signals/src/fixtures/testIndicatorContract.ts
+++ b/packages/trading-signals/src/fixtures/testIndicatorContract.ts
@@ -1,10 +1,15 @@
 import {describe, expect, it} from 'vitest';
+import {NotEnoughDataError} from '../error/NotEnoughDataError.js';
 
-type ReplaceContractOptions<Input> = {
+type IndicatorContractOptions<Input> = {
   /** Creates a fresh instance of the indicator under test */
   create: () => {
     add(input: Input): unknown;
+    getRequiredInputs(): number;
+    getResult(): unknown;
+    getResultOrThrow(): unknown;
     getState(): object;
+    isStable: boolean;
     replace(input: Input): unknown;
   };
   /** Input that must lead the indicator into a different state than the last of `inputs` */
@@ -14,11 +19,36 @@ type ReplaceContractOptions<Input> = {
 };
 
 /**
- * Registers the replace() contract every indicator must fulfil: replacing the latest input
- * with itself changes nothing, and replacing a divergent input restores the exact state of
- * an add-only series.
+ * Registers the contracts every indicator must fulfil. Warm-up: below the declared number of
+ * required inputs there is no result, only a `NotEnoughDataError`. Replace: replacing the
+ * latest input with itself changes nothing, and replacing a divergent input restores the
+ * exact state of an add-only series.
  */
-export function testReplaceContract<Input>({create, divergentInput, inputs}: ReplaceContractOptions<Input>) {
+export function testIndicatorContract<Input>({create, divergentInput, inputs}: IndicatorContractOptions<Input>) {
+  describe('warm-up contract', () => {
+    it('stays unstable and yields no result before the declared warm-up', () => {
+      const indicator = create();
+      const requiredInputs = indicator.getRequiredInputs();
+
+      expect(inputs.length, 'the contract inputs must cover the declared warm-up').toBeGreaterThanOrEqual(
+        requiredInputs
+      );
+
+      const expectNoResult = (count: number) => {
+        expect(indicator.isStable, `unstable after ${count} of ${requiredInputs} required inputs`).toBe(false);
+        expect(indicator.getResult(), `no result after ${count} of ${requiredInputs} required inputs`).toBeNull();
+        expect(() => indicator.getResultOrThrow()).toThrow(NotEnoughDataError);
+      };
+
+      expectNoResult(0);
+
+      inputs.slice(0, requiredInputs - 1).forEach((input, index) => {
+        indicator.add(input);
+        expectNoResult(index + 1);
+      });
+    });
+  });
+
   describe('replace() contract', () => {
     it('changes nothing when the latest input is replaced with itself', () => {
       const indicator = create();
--- a/packages/trading-signals/src/momentum/OBV/OBV.ts
+++ b/packages/trading-signals/src/momentum/OBV/OBV.ts
@@ -35,8 +35,7 @@ export class OBV extends TrendIndicatorSeries<OpenHighLowCloseVolume<number>> {
     const prevPrice = prevCandle.close;
     /*
      * OBV accumulates onto the running total, so a replacement has to build on the total from
-     * before the replaced candle. `this.result` still carries that candle's contribution until
-     * `setResult()` unwinds it, which would count the bar twice.
+     * before the replaced candle.
      */
     const prevResult = (replace ? this.previousResult : this.result) ?? 0;
     const currentPrice = candle.close;
--- a/packages/trading-signals/src/momentum/OBV/OBV.ts
+++ b/packages/trading-signals/src/momentum/OBV/OBV.ts
@@ -25,7 +25,7 @@ export class OBV extends TrendIndicatorSeries<OpenHighLowCloseVolume<number>> {
   }
 
   update(candle: OpenHighLowCloseVolume<number>, replace: boolean) {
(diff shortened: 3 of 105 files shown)


packages/trading-signals/src/volatility/BBANDS/BollingerBands.test.ts
```diff
@@ -153,7 +207,7 @@ describe('BollingerBands', () => {
 
     it('returns BULLISH when price breaks above the previous upper band', () => {
       const bb = new BollingerBands(10, 2);
-      // Need 12 adds: 11 for first result, 12th to have a previous result to compare against
+      // The signal compares against the previous bands, so more than one result must exist
```
bennycode: keep previous comment

packages/trading-signals/src/momentum/ROC/ROC.test.ts
```diff
@@ -31,7 +31,7 @@ describe('ROC', () => {
         }
       });
 
-      expect(roc.getRequiredInputs()).toBe(interval);
+      expect(roc.getRequiredInputs(), 'a comparison needs one bar more than the interval').toBe(interval + 1);
```
bennycode: Testing `roc.getRequiredInputs()` here is the wrong test (as it doesnt relate to the test title).

packages/trading-signals/src/momentum/TDS/TDS.test.ts
```diff
@@ -110,6 +110,112 @@ describe('TDS', () => {
...
+      const replaced = new TDS();
+      closes.slice(0, -1).forEach(close => replaced.add(close));
+      replaced.add(99);
+      replaced.replace(closes[closes.length - 1]);
+
+      const reference = new TDS();
+      closes.forEach(close => reference.add(close));
+
+      expect(replaced.getResult(), 'a replacement must reproduce the add-only series').toBe(reference.getResult());
+      expect(replaced['setupCount']).toBe(reference['setupCount']);
```
bennycode: For indicators where we need to compare a whole state (`setupCount`, `setupDirection`, etc.) and not just one simple return value, I advise to build a comparator. Either as utility in the test OR as an equality in the class itself.

packages/trading-signals/src/trend/ZIGZAG/ZigZag.test.ts
```diff
@@ -43,6 +43,105 @@ describe('ZigZag', () => {
...
+
+    const candles = highs.map((high, index) => ({high, low: lows[index]}));
+
+    function collect(zigzag: ZigZag, inputs: readonly {high: number; low: number}[]) {
+      const results: number[] = [];
+
+      for (const candle of inputs) {
+        const result = zigzag.add(candle);
+
+        if (result !== null) {
```
bennycode: why is checking this important in a `collect` method? Btw. can we conceptually use a batch mechanism here like `updates`? This goals for all indicators where we add values through a for-loop in tests... I think we can simply use `updates` and give them a bunch of data at once.

packages/trading-signals/src/volume/PVT/PVT.test.ts
```diff
@@ -106,6 +106,40 @@ describe('PVT', () => {
   });
 
   describe('replace', () => {
+    it('accumulates onto the total from before the replaced candle', () => {
+      const candle = (close: number) => ({close, high: close, low: close, volume: 1000}) as const;
+
+      const replaced = new PVT();
+      replaced.add(candle(100));
```
bennycode: we can use `updates` here instead of doing 3x `add`

packages/trading-signals/src/volume/PVT/PVT.test.ts
```diff
@@ -106,6 +106,40 @@ describe('PVT', () => {
   });
 
   describe('replace', () => {
+    it('accumulates onto the total from before the replaced candle', () => {
+      const candle = (close: number) => ({close, high: close, low: close, volume: 1000}) as const;
```
bennycode: call it `fakeFlatCandle`. Can we make it a general test util? do we need it somewhere else in tests?

packages/trading-signals/src/volatility/BBW/BollingerBandsWidth.test.ts
```diff
@@ -42,32 +42,39 @@ describe('BollingerBandsWidth', () => {
...
-        '0.21',
-        '0.21',
-        '0.20',
-        '0.18',
-        '0.15',
-        '0.13',
-        '0.11',
-        '0.09',
-        '0.09',
+        '0.18527244', // 0.19
```
bennycode, line 58: we are now emitting one expectation earlier (compared to TradingView). is this correct? is tradingview wrong here?

packages/trading-signals/src/volatility/BBANDS/BollingerBands.test.ts
bennycode: Restored in d4756fa, with one count corrected: the first result now lands on the 10th add (that's what this PR fixes), so it reads "Need 12 adds: 10 for the first result, 12th to have a previous result to compare against". Also restored the comment in 'tracks signal state changes'.

packages/trading-signals/src/momentum/ROC/ROC.test.ts
bennycode: Agreed — moved into its own test under a new `getRequiredInputs` describe block (d4756fa). The tulip test keeps `getRequiredInputs() - 1` only to compute the offset.

packages/trading-signals/src/momentum/TDS/TDS.test.ts
bennycode: Built it as a test utility: `getSetupState()` captures `result`, `setupCount` and `setupDirection`, and both state-comparing tests now assert one `toEqual` on the whole state (d4756fa). Went with the utility over an equality method on the class — the comparison peeks at private fields, which is a test concern, not API surface.

packages/trading-signals/src/trend/ZIGZAG/ZigZag.test.ts
bennycode: The null-check was there because `collect()` returned the emissions — but no caller ever used that return value, so the helper was dead weight. Dropped it entirely; every call site is now `zigzag.updates(...)`, and the one test that does collect emissions uses `zigzag.updates(candles).filter(result => result !== null)` (d4756fa).

Also applied `updates()` across the other test files this PR touches (ROC, TDS, PSAR, PVT, BBANDS) wherever a loop just fed a data array. Left index-generator loops like `for i < 9: add(11 + i)` alone — spelling those out as array literals didn't read better. A repo-wide sweep of untouched test files would be a separate PR.

packages/trading-signals/src/volume/PVT/PVT.test.ts
bennycode: Done — both PVT replace tests now feed their candles through a single `updates()` call (d4756fa).

packages/trading-signals/src/volume/PVT/PVT.test.ts
bennycode: Renamed to `fakeFlatCandle` and hoisted so both replace tests share it (d4756fa). On the general-util question I checked the suite: ULTOSC has a `flatCandle` without volume, MFI's is deliberately not flat (high 51 / low 49). Only PVT needs this exact shape, so a shared util would have exactly one consumer. Kept it local — worth extracting the moment a second same-shape consumer appears.

packages/trading-signals/src/volatility/BBW/BollingerBandsWidth.test.ts
bennycode: TradingView isn't wrong — and after this fix we no longer disagree with it. TV plots its first BBW on the 20th bar (a 20-period SMA needs exactly 20 bars). The old implementation became stable one bar late, so the old test's first checked bar was the 21st and TV's first plotted value was silently skipped — same story as the BB fixture in the PR description.

Verified by recomputing all 12 expectations from the definition `(upper - lower) / middle` with population stddev: the new first value 0.18527244 sits on the 20th bar and rounds to TV's displayed 0.19; every later bar is unchanged from the old list. So the fix moves us onto TV's timing without touching any values. Added a sentence to the test comment documenting this (d4756fa).

packages/trading-signals/src/momentum/TDS/TDS.test.ts
bennycode: Follow-up shipped as draft #1277: `equals()`/`getState()` now live on the indicator base class, backed by a `State` generic and a single state container that `replace()` rewinds generically. The TDS test helper is gone — tests compare via the public API.

packages/trading-signals/src/base/Indicator.ts
```diff
@@ -25,11 +25,41 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
...
+  State extends object = Record<string, never>,
+> implements Indicator<Result, Input> {
   protected result: Result | undefined;
+  /**
+   * Single container for every mutable field (beyond the result) that `update()` may change.
+   * Keeping such fields here instead of in loose class fields lets the base class rewind them
+   * generically on `replace()` (see {@link trackState}) and makes two indicator instances
+   * comparable via {@link getState}.
+   */
+  protected state: State = {} as State;
```
bennycode, line 34: todo: can we remove as assertions here?

packages/trading-signals/src/base/Indicator.ts
```diff
@@ -25,11 +25,41 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
...
  */
-export abstract class TechnicalIndicator<Result, Input> implements Indicator<Result, Input> {
+export abstract class TechnicalIndicator<
+  Result,
+  Input,
+  State extends object = Record<string, never>,
+> implements Indicator<Result, Input> {
   protected result: Result | undefined;
+  /**
+   * Single container for every mutable field (beyond the result) that `update()` may change.
```
bennycode: todo: remove comment

packages/trading-signals/src/fixtures/testIndicatorContract.ts
```diff
@@ -0,0 +1,88 @@
+import {describe, expect, it} from 'vitest';
+import {NotEnoughDataError} from '../error/NotEnoughDataError.js';
+
+type IndicatorContractOptions<Input> = {
+  /** Creates a fresh instance of the indicator under test */
+  create: () => {
+    add(input: Input): unknown;
+    getRequiredInputs(): number;
```
bennycode: can we extract these types from the source than specifying them ourselves?

packages/trading-signals/src/fixtures/testIndicatorContract.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Registers the contracts every indicator must fulfil. Warm-up: below the declared number of
+ * required inputs there is no result, only a `NotEnoughDataError`. Replace: replacing the
+ * latest input with itself changes nothing, and replacing a divergent input restores the
+ * exact state of an add-only series.
+ */
+export function testIndicatorContract<Input>({create, divergentInput, inputs}: IndicatorContractOptions<Input>) {
+  describe('warm-up contract', () => {
+    it('stays unstable and yields no result before the declared warm-up', () => {
```
bennycode, line 22: can we build a test where we test that it is not just unstable until the warm-up finished but also gives a result or at least `null` instead of `undefined` when warmed-up?

packages/trading-signals/src/momentum/OBV/OBV.ts
```diff
@@ -25,19 +25,24 @@ export class OBV extends TrendIndicatorSeries<OpenHighLowCloseVolume<number>> {
...
     if (this.candles.length < this.getRequiredInputs()) {
       return null;
     }
 
     const prevCandle = this.candles[this.candles.length - 2];
     const prevPrice = prevCandle.close;
-    const prevResult = this.result ?? 0;
+    /*
+     * OBV accumulates onto the running total, so a replacement has to build on the total from
+     * before the replaced candle. `this.result` still carries that candle's contribution until
```
bennycode: todo: just keep "OBV accumulates onto the running total, so a replacement has to build on the total from
     * before the replaced candle." but refrain from mentioning concrete code (like `setResult()`) in comments. Save this as a rule of this package (CLAUDE.MD for example).

packages/trading-signals/src/momentum/TDS/TDS.ts
```diff
@@ -1,4 +1,15 @@
-import {TradingSignal, TrendIndicatorSeries} from '../../base/Indicator.js';
+import {TradingSignal, TrendIndicatorSeries, type TradingSignals} from '../../base/Indicator.js';
+
+type TDSState = {
+  closes: number[];
+  /*
```
bennycode: write property comments of types with `/** ... */`, so it renders in VS Code

packages/trading-signals/src/trend/ZIGZAG/ZigZag.ts
```diff
@@ -9,6 +9,21 @@ export type ZigZagConfig = {
...
 
+/**
+ * The swing state {@link ZigZag.update} carries between candles. Held in the base-class state
+ * container so a replacement re-runs the latest candle from the state that candle originally saw.
+ */
+type ZigZagState = {
+  highestExtreme: number | null;
+  isUp: boolean;
+  /*
+   * Whether the latest candle reversed the trend. ZigZag emits only on a reversal and its result
```
bennycode, line 20: remember to write with `/** ... */`

packages/trading-signals/src/base/Indicator.ts
```diff
@@ -25,11 +25,41 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
...
+  #previousState?: State;
 
   abstract getRequiredInputs(): number;
 
+  protected trackState(replace: boolean) {
+    if (replace) {
+      if (this.#previousState !== undefined) {
+        this.state = structuredClone(this.#previousState);
+      }
+    } else {
```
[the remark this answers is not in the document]
bennycode: Fixed in e299124. trackState() now seeds the rollback baseline on a first-ever replace, and PSAR has a regression test comparing two pre-input replacements against a single add via getState().

packages/trading-signals/src/util/pushUpdate.ts
```diff
@@ -1,8 +1,19 @@
...
+  /** Overwrite the last element instead of appending */
+  replace: boolean;
+};
+
 /**
  * Adds an item to the array or replaces the last item in the array.
  * If the array limit size is exceeded, the oldest array element will be removed and returned by the function.
  */
-export function pushUpdate<T>(array: T[], replace: boolean, item: T, maxLength: number) {
+export function pushUpdate<T>({array, item, maxLength, replace}: PushUpdateOptions<T>) {
```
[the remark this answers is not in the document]
bennycode: Accepted as a breaking change: the new signature stays and the break will be called out in the release notes. pushUpdate is an internal helper that leaked into the public surface, so a clean signature beats a compatibility overload here.

packages/trading-signals/src/base/Indicator.ts
bennycode: Looked into removing it: an optional field degrades getState() to Partial and breaks typed consumers, a definite-assignment assertion makes the same unchecked claim in different syntax, and injecting initial state adds a constructor parameter to every indicator. The assertion is the initialized-by-subclass idiom, so it stays.

packages/trading-signals/src/base/Indicator.ts
bennycode: Removed in 262d079.

packages/trading-signals/src/fixtures/testIndicatorContract.ts
bennycode: Done in 262d079: create is now typed as () => TechnicalIndicator<unknown, Input, object>, so the contract follows the base class instead of restating it.

packages/trading-signals/src/fixtures/testIndicatorContract.ts
bennycode: Added in 262d079: after the contract inputs the indicator must be stable, getResult() must not return undefined, and getResultOrThrow() must yield an actual result. All 61 wired suites pass it.

packages/trading-signals/src/momentum/OBV/OBV.ts
bennycode: Trimmed in 262d079 — PVT carried the same wording and was trimmed too. The convention is recorded in the package CLAUDE.md.

packages/trading-signals/src/momentum/TDS/TDS.ts
bennycode: Done in 262d079.

packages/trading-signals/src/trend/ZIGZAG/ZigZag.ts
bennycode: Done in 262d079.

PR #1277, opened 2026-08-03
feat(trading-signals): compare indicator state via getState()

bennycode:
# Good design decisions remove a whole class of bugs.

Stacked on #1261 — retarget to `main` once that merges. Follows up on [this review thread](https://github.com/bennycode/trading-signals/pull/1261#discussion_r3703885704): instead of a test-local comparator, internal state is now comparable on the base class.

## What

`TechnicalIndicator` gains a third generic `State` and a single mutable-state container:

- **`protected state: State`** — every mutable field `update()` may touch (beyond the result) lives in one object.
- **`trackState(replace)`** — snapshots `state` before each add, rewinds it on replace. One call at the top of `update()` replaces all per-field rollback bookkeeping.
- **`getState()`** — defensive snapshot of `result` + `state`. Tests assert `expect(a.getState()).toEqual(b.getState())` and get a real diff on failure.

Migrated: **PVT, TDS, PSAR, ZigZag** — the four state-machine indicators fixed in #1261. Their hand-rolled snapshot/restore code (PSAR's `captureState`/`restoreState`, TDS's three shadow fields, ZigZag's inline state copy) is deleted; the `replace()` semantics are unchanged and the whole suite passes untouched except for private-field access moving to `getState()`.

## Why the container instead of a hand-maintained `getState()`

TypeScript cannot reflect over private fields, so a hand-written state snapshot can silently omit a field — the same failure mode as the forgotten rollbacks #1261 fixed. With the container, completeness holds by construction: a field in `state` is rewound and compared automatically; a field outside it sticks out in review.

## Design notes

- `State` is appended **after** existing type parameters on all three base classes — existing subclasses (internal and external) compile unchanged.
- Comparing two indicators is left to `getState()` plus the test framework's own deep equality, which reports a diff on failure. No `equals()` helper on the base class ([review feedback](https://github.com/bennycode/trading-signals/pull/1277#discussion_r3710235735)).
- Indicators not yet migrated expose their result alone. Migration of the remaining stateful indicators (smoothing-based ones like EMA/ATR/ADX and window-based ones) can follow incrementally; end state is that `update()` becomes a base-class template method and the `replace` flag disappears from indicator logic entirely.
- Snapshots use `structuredClone` (Node ≥ 17, all evergreen browsers). Per-add cost is a clone of a small plain-data object (ZigZag: 4 scalars, TDS: ≤ 13 closes, PSAR: 6 fields, PVT: 2 candles).
- Result withdrawal for sparse emitters (TDS setups, ZigZag pivots) keeps using `rollbackLastResult()`; the emitted-flag now lives in `state` and is read before `trackState()` rewinds it.

## Verification

- 564 tests pass, coverage stays at 100% (statements, branches, functions, lines)
- `tsc --noEmit` clean, oxlint/oxfmt clean
- `trading-strategies` unchanged: compiles and passes all 259 tests
- All #1261 replace-correctness sweeps (same-value no-op at every series length, add-only equivalence with decoy bars) pass against the rewritten indicators

### DIFF
--- a/packages/trading-signals/src/base/Indicator.ts
+++ b/packages/trading-signals/src/base/Indicator.ts
@@ -42,12 +42,6 @@ export abstract class TechnicalIndicator<
 
   abstract getRequiredInputs(): number;
 
-  /**
-   * Rewinds `state` by one bar on a replacement, or snapshots it before a regular add. Stateful
-   * indicators call this first in `update()`, so a replacement re-runs the latest bar from the
-   * exact state that bar originally saw — without per-field rollback bookkeeping, which is easy
-   * to get wrong when a field is forgotten.
-   */
   protected trackState(replace: boolean) {
     if (replace) {
       if (this.#previousState !== undefined) {
@@ -59,9 +53,8 @@ export abstract class TechnicalIndicator<
   }
 
   /**
-   * Snapshot of the result and the declared mutable state. Two indicators fed equivalent input
-   * series must produce identical snapshots, which makes state divergence visible as a diff in
-   * tests: `expect(a.getState()).toEqual(b.getState())`.
+   * Snapshot of the result and the mutable state, so two indicator instances fed the same
+   * inputs can be compared.
    */
   getState() {
     return structuredClone({...this.state, result: this.result});
--- a/packages/trading-signals/src/base/Indicator.ts
+++ b/packages/trading-signals/src/base/Indicator.ts
@@ -25,10 +25,47 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
 /**
  * Implements common update behaviour among indicators.
  */
-export abstract class TechnicalIndicator<Result, Input> implements Indicator<Result, Input> {
+export abstract class TechnicalIndicator<
+  Result,
+  Input,
+  State extends object = Record<string, never>,
+> implements Indicator<Result, Input> {
   protected result: Result | undefined;
+  /**
+   * Single container for every mutable field (beyond the result) that `update()` may change.
+   * Keeping such fields here instead of in loose class fields lets the base class rewind them
+   * generically on `replace()` (see {@link trackState}) and makes two indicator instances
+   * comparable via {@link getState}.
+   */
+  protected state: State = {} as State;
+  #previousState?: State;
 
   abstract getRequiredInputs(): number;
+
+  /**
+   * Rewinds `state` by one bar on a replacement, or snapshots it before a regular add. Stateful
+   * indicators call this first in `update()`, so a replacement re-runs the latest bar from the
+   * exact state that bar originally saw — without per-field rollback bookkeeping, which is easy
+   * to get wrong when a field is forgotten.
+   */
+  protected trackState(replace: boolean) {
+    if (replace) {
+      if (this.#previousState !== undefined) {
+        this.state = structuredClone(this.#previousState);
+      }
+    } else {
+      this.#previousState = structuredClone(this.state);
+    }
+  }
+
+  /**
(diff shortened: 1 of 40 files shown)


packages/trading-signals/src/base/Indicator.ts
```diff
@@ -25,11 +25,58 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
...
+    return structuredClone({...this.state, result: this.result});
+  }
+
+  /**
+   * `true` when both indicators hold identical internal state, i.e. they will produce identical
+   * results for identical future inputs. Only the result and fields kept in `state` take part in
+   * the comparison — an indicator that still keeps mutable fields outside of `state` is compared
+   * by its result alone.
+   */
+  equals(other: TechnicalIndicator<Result, Input, State>): boolean {
```
bennycode: @claude remove the `equals` functionalty and leave the PR just to `getState()`

packages/trading-signals/src/base/Indicator.ts
```diff
@@ -25,11 +25,48 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
...
+      if (this.#previousState !== undefined) {
+        this.state = structuredClone(this.#previousState);
+      }
+    } else {
+      this.#previousState = structuredClone(this.state);
+    }
+  }
+
+  /**
+   * Snapshot of the result and the declared mutable state. Two indicators fed equivalent input
```
bennycode: todo: remove the technical details, don't use demo code in code comments

packages/trading-signals/src/base/Indicator.ts
```diff
@@ -25,11 +25,48 @@ export type TradingSignals = (typeof TradingSignal)[keyof typeof TradingSignal];
...
+   * generically on `replace()` (see {@link trackState}) and makes two indicator instances
+   * comparable via {@link getState}.
+   */
+  protected state: State = {} as State;
+  #previousState?: State;
 
   abstract getRequiredInputs(): number;
 
+  /**
+   * Rewinds `state` by one bar on a replacement, or snapshots it before a regular add. Stateful
```
bennycode: todo: remove comment, functionality is expressed in tests

packages/trading-signals/src/trend/ZIGZAG/ZigZag.ts
```diff
@@ -10,12 +10,17 @@ export type ZigZagConfig = {
...
+ * container so a replacement re-runs the latest candle from the state that candle originally saw.
  */
 type ZigZagState = {
   highestExtreme: number | null;
   isUp: boolean;
+  /*
+   * Whether the latest candle reversed the trend. ZigZag emits only on a reversal and its result
+   * persists in between, so only a candle that emitted may withdraw that pivot when replaced.
+   */
+  lastCandleReversed: boolean;
```
bennycode: todo: name it `lastCandleReversedTrend`

packages/trading-signals/src/volume/PVT/PVT.ts
```diff
@@ -15,21 +19,24 @@ import {pushUpdate} from '../../util/pushUpdate.js';
...
   override getRequiredInputs() {
     return 2;
   }
 
   update(candle: HighLowCloseVolume, replace: boolean) {
-    pushUpdate(this.#candles, replace, candle, 2);
+    this.trackState(replace);
+
+    // trackState() already rewound the window on a replacement, so the candle is always appended
+    pushUpdate(this.state.candles, false, candle, 2);
```
bennycode: todo: pushUpdate has now 4 params, give it an object so we cannot accidentally mix up params.

PR #1279, opened 2026-08-03
feat(exchange): configurable slippage in BrokerMock

### LINKED ISSUE #1199: feat(exchange): configurable slippage in BrokerMock
## Problem

`BrokerMock` fills market orders at the candle **open** with zero slippage, and limit orders with price improvement (`min(limit, open)` for buys / `max(limit, open)` for sells). Real fills are worse than that — backtest results are systematically optimistic, especially for high-frequency strategies like `ScalpStrategy` where per-fill slippage dominates the edge.

## Proposal

Add an optional slippage model to `BrokerMock` / `BacktestConfig`:

- **Market orders**: fill at `open * (1 ± slippage)` (against the order direction). A single configurable rate (e.g. `0.05%`) is enough for v1.
- **Limit orders**: option to disable price improvement (fill at the limit price, not better) — the conservative assumption.
- **Default: 0** — existing backtests stay reproducible; slippage is opt-in.

## Notes

- Keep it in the matching engine (`BrokerMock.#tryMatch`) next to the existing fee deduction, so fee + slippage math stays in one place.
- Expose the setting in the docs backtester config as a follow-up.
- Later extension: bid/ask-spread-based slippage once candles carry ask data (the `Candle` schema already has optional ask fields).

Sahilhawal:
Right now `BrokerMock` fills market orders at exactly the candle open with zero slippage, and limit orders always get price improvement (min/max against the open). That's unrealistic — real fills are worse, so backtests end up looking better than they'd actually perform, especially for fast strategies where every bit of slippage matters.

This adds an optional `slippage` config to `BrokerMock`:

- `rate` — applied against market fills in the direction that hurts you (buys fill a bit higher, sells fill a bit lower)
- `disablePriceImprovement` — makes limit orders fill at the limit price instead of getting the free price improvement toward the open

Both default to off, so existing backtests don't change unless you opt in. Wired it through `AlpacaBrokerMock` the same way `feeRates` already is.

Closes #1199.

Tested with a new `slippage` test block in `BrokerMock.test.ts` (buy/sell slippage, default-off, and disabled price improvement on both sides), plus the full exchange + trading-strategies suites and a full build to make sure nothing else broke.

### DIFF
--- a/packages/exchange/src/broker/BrokerMock.test.ts
+++ b/packages/exchange/src/broker/BrokerMock.test.ts
@@ -1,6 +1,11 @@
 import Big from 'big.js';
 import {describe, expect, it} from 'vitest';
-import {BrokerMock, type ExchangeMockBalance} from './BrokerMock.js';
+import {
+  BrokerMock,
+  type BrokerMockFillModelConfig,
+  type BrokerMockSlippageConfig,
+  type ExchangeMockBalance,
+} from './BrokerMock.js';
 import {type Candle, type FeeRate, type Fill, OrderSide, OrderType, type TradingRules} from './Broker.js';
 import {TradingPair} from './TradingPair.js';
 import {ms} from 'ms';
@@ -19,9 +24,14 @@ class TestExchangeMock extends BrokerMock {
     counter_min_size: '1',
   };
 
-  constructor(balances: Map<string, ExchangeMockBalance>) {
-    super({balances});
+  constructor(
+    balances: Map<string, ExchangeMockBalance>,
+    slippage?: BrokerMockSlippageConfig,
+    fillModel?: BrokerMockFillModelConfig
+  ) {
+    super({balances, fillModel, slippage});
     this.setCachedFeeRates(TestExchangeMock.TEST_FEE_RATES);
+    this.setCachedTradingRules(TestExchangeMock.TEST_TRADING_RULES);
   }
 
   async getFeeRates(): Promise<FeeRate> {
@@ -60,12 +70,19 @@ function createCandle(overrides: Partial<Candle> & {open: string; close: string}
   };
 }
 
-function createExchange(baseAmount: string, counterAmount: string) {
+function createExchange(
+  baseAmount: string,
+  counterAmount: string,
+  slippage?: BrokerMockSlippageConfig,
+  fillModel?: BrokerMockFillModelConfig
+) {
   return new TestExchangeMock(
     new Map([
       ['BTC', {available: new Big(baseAmount), hold: new Big(0)}],
       ['USD', {available: new Big(counterAmount), hold: new Big(0)}],
-    ])
+    ]),
+    slippage,
+    fillModel
   );
 }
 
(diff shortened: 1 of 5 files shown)


Sahilhawal:
@bennycode can you take a look please? This adds optional slippage to `BrokerMock` for #1199 — should be a safe opt-in since it defaults to off.

bennycode (commented):
Market-order slippage is the right idea and default-off is the right call. Two blockers below, plus the feature currently can't be reached from the backtest CLI.

## `disablePriceImprovement` fills at prices that never traded

Both new tests assert impossible fills:

```
limit buy  @ 500, candle open 360 / high 400 / low 350  -> fill 500
limit sell @ 400, candle open 450 / high 460 / low 420  -> fill 400
```

Nothing traded at 500 in a candle whose high was 400. 

## Fill price ignores `counter_increment`

`105 * 1.0033 = 105.3465` is off-tick. `placeOrder` already runs limit prices through `#roundDownToIncrement`. Slipped market fills need the same, rounded against the trader (up on buys, down on sells).

## No validation on `rate`

A negative rate silently gives buys price improvement, which is the bug this PR exists to fix. `rate: 1` fills every sell at 0.

```ts
assert.ok(
  this.#slippageRate.gte(0) && this.#slippageRate.lt(1),
  `Slippage rate "${this.#slippageRate.toFixed()}" must be within [0, 1)`
);
```

## Type `rate` as `Big` and document the scale

`FeeRate` is the same concept one file over:

```ts
/** A fee rate of "1" means 100%, "0.1" is 10%, "0.01" is 1%, ... */
MARKET: Big;
```

`rate?: Big | number | string` with no doc leaves `rate: 5` ambiguous between 5%, 5 bps and 500%. Should be:

```ts
export interface BrokerMockSlippageConfig {
  /** Fraction of the fill price lost to slippage: "0.01" is 1%. Applied against market fills only. */
  rate?: Big;
}
```

Also, `disablePriceImprovement` isn't slippage. If it stays, it belongs in a `fillModel` config, phrased positively as `priceImprovement?: boolean` defaulting to `true`.

## Hold estimate excludes slippage

Market buys hold `size * currentCandle.close * (1 + fee)` but now fill at `nextOpen * (1 + slippage)`. `#applyFill` settles the shortfall through `#addAvailable`, which has no balance floor, so `available` can go negative. Pre-existing on upward gaps, but slippage makes it systematic on every buy. Pad the hold by the slippage rate.

## Not reachable from `runBacktest`

`packages/trading-strategies/src/start/runBacktest.ts` builds `AlpacaBrokerMock` with no slippage and exposes no flag, so the feature only exists in test code. Add `--slippage` next to `--balance`.

## Missing tests

- `sizeInCounter: true` market buy, where `fillPrice` feeds back into `netCounter.div(fillPrice)`. Most interesting path in the diff, untested.
- Slippage does *not* apply to limit orders. Deliberate decision, should be executable.
- `rate` and `disablePriceImprovement` combined.
- `AlpacaBrokerMock` actually threading the option through. That's the only production wiring here.
- Rejection of an invalid rate, once validation exists.

Assertion comments go in vitest's message argument:

```ts
expect(fills[0].price, '105 open + 1% slippage').toBe('106.05');
```

## Nits

Nested ternary, duplicated on both branches. Hoist it above the buy/sell split:

```ts
if (this.#disablePriceImprovement) {
  fillPrice = limitPrice;
} else if (order.side === OrderSide.BUY) {
  ...
}
```

Stale comments the diff should have removed: `// Market orders fill at candle open price`, and both `// Price improvement: min/max(order.price, candle.open)`. They restate the code, so delete rather than update.

bennycode (changes requested):

Sahilhawal:
@bennycode I have made the changes, could you please take a look?

bennycode:
@Sahilhawal can you tell me why you would need this feature? It would actually only affect market orders in a backtesting environment.

Sahilhawal:
@bennycode Need this because right now market orders in BrokerMock fill at the candle open with zero cost, so backtests come out looking better than real trading would be. It matters most for something like ScalpStrategy that fires a lot of market orders, each trade only has a small edge, so slippage can eat all of it or flip it negative. Wanted a way to catch that before putting real money behind a strategy. Scope is narrow on purpose too, same as what you wrote in the spec on #1199, slippage only touches market orders, limit orders aren't affected by rate, and it defaults to off so nothing changes unless someone turns it on.

On the review from the 3rd, these are all done now:

- Impossible fills from disablePriceImprovement, clamped to candle high/low instead of the raw limit price
- Slipped market fills round to the counter increment
- rate validated to [0, 1) and typed as Big
- disablePriceImprovement renamed to fillModel.priceImprovement, defaults to true
- Hold estimate padded for slippage so available balance can't go negative
- --slippage flag in runBacktest.ts (this one was actually already there from the first commit)
- Added the tests you asked for: notional buy with slippage, limit orders not affected, rate and priceImprovement combined, AlpacaBrokerMock wiring, invalid rate getting rejected

I actually just saw your spec comment on the issue today, and noticed one part isn't handled yet, it says slippage should only happen within a candle's range, low to high. That part isn't done for market orders, the fill price isn't clamped to candle high/low like the limit order fix is, so on a tight candle a big rate could still produce a fill that never actually happened. Still open, working on it.

let me know if I am missing anything

Sahilhawal:
Hey @bennycode shall I close this PR

bennycode:
Thanks for leaving maintainer edits enabled, that made this much easier. I pushed straight to your branch instead of sending you another round of review notes. Sorry for the long silence before that.

I turned it into a simple slippage config. The whole surface is now:

```ts
slippage: {
  rate?: Big;       // "0.01" is 1%, market orders only
  clamp?: boolean;  // keep fills inside the candle's traded range, default true
}
```

What I removed:

- `fillModel.priceImprovement`. It's a separate fill-model change rather than slippage. Worth its own PR and I'd take it, see below.
- Tick rounding of slipped fills, and with it `setCachedTradingRules()`. That was public API on `BrokerMock` that threw at runtime if a subclass forgot to call it. Off-tick fill prices are cosmetic since all the arithmetic stays in `Big`.
- The slippage padding on the market-buy hold. That hold has always been an estimate an upward gap can exceed, so fixing it is its own change.

To be clear about why this shrank: most of those were things I asked you for in my review on the 3rd, and you delivered them. The scope was wrong on my end, not your execution.

The one thing I added is `clamp`. Your instinct to clamp fills to the candle range was right, but once it was in I measured it, and the clamp caps the effective rate at the range itself. On our scalp backtest the rate stops doing anything above roughly 5 basis points:

```
rate=0.0001  P&L 479.95
rate=0.0005  P&L 476.47
rate=0.0050  P&L 476.47   <- saturated
```

The sharper version of the same problem: for a buy the fill is `min(open * (1 + rate), high)`, so on any candle whose high equals its open, a buy pays nothing at any rate. Sells mirror that on candles that open at their low. Slippage ends up conditional on which way the candle went after the open, which real slippage isn't.

backtrader hits the same wall and solves it with the `slip_match` / `slip_out` pair, so I kept your clamp as the default and made it switchable. On the backtest CLI that's `--slippage-rate` and `--no-slippage-clamp`.

Not closing this. Your commit is still the base of the branch.

On `priceImprovement`, it's worth its own PR and I'll review it properly this time. Right now market orders are modelled pessimistically while limit orders still take the favourable side of the open on every fill, so the two order types carry opposite biases. Your flag is the fix for that half.

bennycode:
Hey @Sahilhawal, thanks for sticking with this!

I modified your PR (thanks for allowing that) and added `--slippage-rate` and `--no-slippage-clamp` to the Backtesting CLI. 🥳

Sahilhawal:
Hello @bennycode,

Definitely! It is your repo after all 😄

Thanks a lot for the edits.

Also, just wanted to say thanks for building and maintaining `trading-signals`. It is genuinely a great library. I have used it in one of my production apps as well as a side project, and it has saved me a ton of time. Happy to keep contributing wherever I can!

PR #1281, opened 2026-08-05
refactor(trading-strategies): unify strategy init() contract + enforce no-unsafe-assignment

bennycode:
## Summary

Two related type-safety changes:

### 1. Unified `init()` warmup contract

The optional strategy `init()` hook had drifted into two incompatible signatures: the `TradingSessionStrategy` interface declared `init(market, pair)` while `ScalpStrategy` implemented `init(candles)` — a live `TradingSession` would have crashed calling it. Nothing caught this because the `implements` check only runs on the base class (where the optional member is simply absent) and both call sites bypassed the type (upcast + duck-typing).

- `Strategy` now provides a concrete **no-op `init(market, pair)` hook**, so the member participates in the `implements TradingSessionStrategy` check, every subclass override is compile-checked against the base signature (a drifted override fails the build — verified with a probe class), and strategies without a warmup need implement nothing
- `ScalpStrategy` overrides it and fetches its own warmup history (30 daily candles for the ATR offset, ER filter, and EMA seeding)
- `runBacktest` drops the `'init' in strategy` duck-typing and calls through the typed interface via a candle-array-backed `MarketDataSource` stub that serves the file's full history — preserving existing warmup semantics (AMD case-study backtest verified bit-identical to `main`: offset $2.04, 30 trades, 18.19%)
- Candle files are now zod-validated (`z.array(CandleSchema)`) and strategy configs typed `unknown` instead of leaking `any` from `JSON.parse`

### 2. `typescript/no-unsafe-assignment` enforced repo-wide

Added next to the existing `no-unsafe-argument` and fixed all 59 violations (exchange 5, messaging 17, trading-signals-docs 32, trading-strategies 5) — no `as any`, no disable comments:

- **Test mocks**: untyped `vi.fn()` was the root cause of most violations; mocks now carry real signatures, which made several assertions strictly stronger
- **Boundaries**: `JSON.parse` results and WebSocket frames land in `unknown` + zod/type-guard validation
- **Docs**: one shared typed Highcharts tooltip formatter replaces eight copy-pasted `(this as any).points` blocks; indicator demo plumbing loses its `TIndicator = any` defaults

### Infrastructure fix included

Type-aware linting was **silently dead in local root runs**: oxlint's nested-config discovery picked up the `oxlint.config.ts` copies inside `.claude/worktrees/agent-*`, rejected their `typeAware` option as non-root, and aborted the entire type-aware pass without failing the build (CI was unaffected — fresh checkouts have no worktrees). Gitignoring `.claude/worktrees/` fixes this — oxlint's discovery respects `.gitignore` — and also unbreaks `oxfmt` locally (128 phantom formatting diffs). Verified with a planted-violation probe that the type-aware pass now runs from the repo root.

## Test plan

- [x] `npm run lint` exit 0 repo-wide; `no-unsafe-assignment` count: 0
- [x] Planted-violation probe: bare root `oxlint` run catches a fresh unsafe assignment (type-aware pass confirmed live)
- [x] `tsc --noEmit` clean in exchange, messaging, trading-strategies, trading-signals-docs
- [x] All unit tests green: exchange 89, messaging 84, trading-strategies 261
- [x] Drift probe: old-style `init(candles)` override fails compilation
- [x] AMD scalp backtest output bit-identical to `main`


packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -81,6 +71,16 @@ export abstract class Strategy implements TradingSessionStrategy {
     }
   }
 
+  /**
+   * Warm-up hook, invoked by the runtime (live `TradingSession` or the backtest runner)
```
bennycode: todo: comment is too exhaustive

packages/trading-strategies/src/start/runBacktest.ts
```diff
@@ -90,32 +90,29 @@ const exchange = new AlpacaBrokerMock({
...
-    console.log(`Auto-computed offset: ${strategy.config.offset} ${counter}`);
-  }
-
-  if (strategy instanceof ScalpStrategy) {
-    console.log(
-      `Scalp-friendly (ER): ${strategy.scalpFriendly ? 'Yes' : 'No — stock is trending, strategy will not trade'}`
-    );
-  }
+/*
+ * 4. Warm up the strategy with the backtest file's own history, via a candle-array-backed
```
bennycode: todo: can the backtest in such a case use a market data source and fetch live from a market data source the candles that would be necessary?

packages/trading-strategies/src/strategy-scalp/ScalpStrategy.ts
```diff
@@ -55,27 +62,27 @@ export class ScalpStrategy extends ProtectedStrategy {
...
-   * Pre-seed the EMA with historical candles to skip the live warmup period.
-   * When no offset is configured, it is auto-computed from these candles using
-   * daily ATR (takes <5ms even on 20k+ candles).
+   * Pre-seeds the EMA with recent history to skip the live warmup period. When no offset
+   * is configured, it is auto-computed from daily ATR (takes <5ms even on 20k+ candles).
    *
-   * Also evaluates Range Efficiency (ER) to determine if the stock is
-   * scalp-friendly. Trending stocks (ER >= 0.4) are flagged as unsuitable.
+   * Also evaluates Range Efficiency (ER) to determine if the stock is scalp-friendly.
+   * Trending stocks (ER at or above {@link ScalpStrategy.ER_THRESHOLD}) are flagged as unsuitable.
```
bennycode: todo: try to use less technical terms and constants bound to code in comments

packages/trading-strategies/src/strategy/Strategy.ts
bennycode: Trimmed to two lines in 877c2af — just what the hook is for and that the no-op is the opt-out.

packages/trading-strategies/src/start/runBacktest.ts
bennycode: Yes — implemented in 877c2af. With Alpaca credentials in the exchange package's .env, the warmup now fetches candles live from just before the backtest window (so it never sees the replayed candles — no more look-ahead). Without credentials, or when the fetch fails (e.g. non-Alpaca pairs like BNB/USDT), it falls back to the file's own history as before. Note: with credentials configured, auto-computed offsets can differ slightly from the README case-study numbers, since the warmup window no longer overlaps the backtest itself.

packages/trading-signals-docs/pages/backtest.tsx
```diff
@@ -60,6 +60,10 @@ function createStrategy(strategyId: StrategyId, config: Record<string, unknown>)
   }
 }
 
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null;
+}
```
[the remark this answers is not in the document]
bennycode: Good catch — fixed in 877c2af: isRecord() now excludes arrays, so an array config can no longer be spread into the protection object.

packages/trading-strategies/src/strategy-scalp/ScalpStrategy.ts
bennycode: Reworded in 877c2af — plain language, no constants or indicator jargon in the warm-up comments.

.claude/rules/vitest-mocking.md
```diff
@@ -1,5 +1,22 @@
...
+const getBalances = vi.fn().mockResolvedValue({basee: new Big(1)});
+
+// ✅ Good: signature makes the payload compiler-checked
+const getBalances = vi.fn<AlpacaBroker['getAvailableBalances']>().mockResolvedValue({
+  base: new Big(1),
+  counter: new Big(5000),
+});
+```
+
+Derive the signature instead of hand-writing it: `SomeType['method']`, `typeof SomeClass.staticMethod`, or `NonNullable<SomeType['optionalMethod']>`. A bare `vi.fn()` remains fine for pure call-recording sinks (`toHaveBeenCalledTimes` and friends) where no data flows in either direction.
```
bennycode: todo: "and friends" is a bit loose writing

packages/messaging/src/service/StrategyMonitor.ts
```diff
@@ -14,6 +15,12 @@ interface ActiveSession {
   strategy: TradingStrategy;
 }
 
+/**
+ * Guards the DB → runtime boundary: `row.state` is a JSON string written by an earlier process,
```
bennycode: todo: remove comment

packages/trading-signals-docs/components/Chart.tsx
```diff
@@ -21,6 +21,24 @@ export interface ChartProps {
   flags?: FlagPoint[];
 }
 
+type TooltipFormatter = NonNullable<NonNullable<ChartOptions['tooltip']>['formatter']>;
+
+/**
+ * Every multi-series demo chart shares one tooltip look; `formatValue` covers the per-chart
```
bennycode: todo: remove technical words like `formatValue`

packages/trading-signals-docs/utils/processData.ts
```diff
@@ -11,11 +11,19 @@ interface ProcessDataOptions {
   alwaysStable?: boolean;
 }
 
+/** Narrower than `DemoIndicator`: every indicator routed through here yields a single numeric result. */
+interface SeriesIndicator {
```
bennycode: todo: cant we reuse the types from the "trading-signals" module?

packages/trading-signals-docs/utils/types.ts
```diff
@@ -1,15 +1,34 @@
...
+  render?: (val: unknown, row?: Record<string, unknown>) => ReactNode;
   className?: string;
 }
 
-interface BaseIndicatorConfig<TIndicator = any> {
+/**
+ * What the generic demo plumbing relies on from a `trading-signals` indicator. Result and input
+ * shapes differ per indicator, so a config that reads them parameterizes `IndicatorConfig<TIndicator>`.
+ */
+export interface DemoIndicator {
```
bennycode: reuse trading-signals type. you can use extract or other TypeScript utility types to achieve this.

packages/trading-strategies/src/start/runBacktest.ts
```diff
@@ -1,10 +1,21 @@
...
 import {createStrategy, getStrategyNames} from '../strategy/StrategyRegistry.js';
+import {ScalpStrategy} from '../strategy-scalp/ScalpStrategy.js';
+
+/*
+ * The exchange package owns the credentials. Load its env so this script can run from
+ * trading-strategies/ without duplicating secrets. Its defaults file is deliberately not
+ * loaded — the placeholder values in there would masquerade as real credentials and trigger
+ * doomed live-warmup requests on machines without a configured .env.
+ */
+config({path: '../exchange/.env'});
```
bennycode, line 20: comment: use .env from the strategies... we could actually think about having only one .env on the root of the monorepo, would that work?

packages/trading-strategies/src/strategy-scalp/ScalpStrategy.test.ts
```diff
@@ -284,7 +290,39 @@ describe('ScalpStrategy', () => {
       }
     }
 
-    // Should not throw — offset is auto-computed from the candles
-    strategy.init(candles.map(toBatched));
+    await strategy.init(marketWith(candles), pair);
+
+    // ATR(14) * 0.2 over the fixture's daily aggregation
```
bennycode: todo: dont write comments above expectations, use them to fill out expectation messages (second param)

packages/trading-strategies/src/strategy/Strategy.ts
```diff
@@ -71,6 +71,12 @@ export abstract class Strategy implements TradingSessionStrategy {
     }
   }
 
+  /**
+   * Warm-up hook, called before the first candle so a strategy can fetch history and
```
bennycode: todo: remove comment

bennycode: Removed in 528fb94.

packages/messaging/src/service/StrategyMonitor.ts
bennycode: Removed in 528fb94.

.claude/rules/vitest-mocking.md
bennycode: Reworded in 528fb94.

packages/trading-signals-docs/components/Chart.tsx
bennycode: Reworded in 528fb94.

packages/trading-strategies/src/strategy-scalp/ScalpStrategy.test.ts
bennycode: Done in 528fb94.

packages/trading-signals-docs/utils/processData.ts
bennycode: Done in 528fb94. The types now derive from IndicatorSeries/TechnicalIndicator via Pick, and the signal shape via ReturnType. add stays a locally declared method: mapped types turn methods into properties, which would break assignability for indicators with narrower input types.

packages/trading-signals-docs/utils/types.ts
bennycode: Done in 528fb94. The types now derive from IndicatorSeries/TechnicalIndicator via Pick, and the signal shape via ReturnType. add stays a locally declared method: mapped types turn methods into properties, which would break assignability for indicators with narrower input types.

packages/trading-strategies/src/start/runBacktest.ts
bennycode: A single root .env works. dotenv never overwrites keys that are already set, so the script now loads ../../.env first and the exchange .env second. Once secrets move to the root file, the exchange lookup finds nothing and can be deleted. runStrategy and the messaging service still read the exchange .env, so a full consolidation means updating those two call sites.

PR #1287, opened 2026-08-11
fix(trading-signals): re-seed EMA, RMA and WSMA when the seeding price is replaced

bennycode:
## Summary

Replacing the price that seeded one of the recursive moving averages produced a corrupted reading instead of re-seeding:

- **EMA / RMA**: `replace()` on the very first price fell through to the non-replace formula and blended the new price with the stale seed (`price·k + oldSeed·(1−k)`). Example: `EMA(5)` seeded with 10, `replace(20)` returned 13.33 instead of 20.
- **WSMA**: `replace()` on the price that completed the first window matched no branch at all — it returned `null` and silently kept the stale result, even though the internal SMA window had already been updated.

## Fix

All three now derive the continuation value the same way: the reading committed before the incoming price (`previousResult` on replace, current result on add), falling back to a fresh seed when no such reading exists. Add-only computation paths are bit-for-bit unchanged — the full suite (763 tests, incl. all Tulip-verified expectations) passes untouched.

## Tests

One regression test per class proving that replacing the seed price reproduces an add-only series **exactly** (`toBe`, no tolerance):

- `EMA`: add sentinel → replace with real first price → feed remainder → bitwise-equal to add-only instance
- `RMA`: same shape
- `WSMA`: replace the window-completing price → continue → bitwise-equal

## Impact

Every indicator composing these averages (DEMA, TEMA, TRIX, MACD, PPO, HMA, ATR via WSMA, …) now handles a replace at its seeding boundary correctly. Found while implementing KVO in #1285 (which carries its own recursion for Tulip seeding parity and is unaffected).

### DIFF
--- a/packages/trading-signals/src/trend/EMA/EMA.test.ts
+++ b/packages/trading-signals/src/trend/EMA/EMA.test.ts
@@ -28,15 +28,17 @@ describe('EMA', () => {
       const interval = 5;
       const ema = new EMA(interval);
       const emaWithReplace = new EMA(interval);
+      const firstPrice = prices[0];
+      const remainingPrices = prices.slice(1);
 
       for (const price of prices) {
         ema.add(price);
       }
 
       emaWithReplace.add(90210);
-      emaWithReplace.replace(prices[0]);
+      emaWithReplace.replace(firstPrice);
 
-      for (const price of prices.slice(1)) {
+      for (const price of remainingPrices) {
         emaWithReplace.add(price);
       }
 
--- a/packages/trading-signals/src/trend/EMA/EMA.test.ts
+++ b/packages/trading-signals/src/trend/EMA/EMA.test.ts
@@ -29,11 +29,16 @@ describe('EMA', () => {
       const ema = new EMA(interval);
       const emaWithReplace = new EMA(interval);
 
-      ema.updates(prices, false);
+      for (const price of prices) {
+        ema.add(price);
+      }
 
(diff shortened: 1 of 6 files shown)


packages/trading-signals/src/trend/EMA/EMA.test.ts
```diff
@@ -24,6 +24,25 @@ describe('EMA', () => {
...
+      const interval = 5;
+      const ema = new EMA(interval);
+      const emaWithReplace = new EMA(interval);
+
+      for (const price of prices) {
+        ema.add(price);
+      }
+
+      emaWithReplace.add(90210);
+      emaWithReplace.replace(prices[0]);
```
bennycode: add variables for `firstPrice = prices[0]` and `remainingPrices = prices.slice(1)` to make the test more readable

packages/trading-signals/src/trend/RMA/RMA.test.ts
```diff
@@ -20,6 +20,26 @@ describe('RMA', () => {
...
+      const inputPrices = [11, 12, 13, 14, 15, 16] as const;
+
+      for (const price of inputPrices) {
+        rma.add(price);
+      }
+
+      rmaWithReplace.add(90210);
+      rmaWithReplace.replace(inputPrices[0]);
+
+      for (const price of inputPrices.slice(1)) {
```
bennycode: see comment at `EMA`

PR #1315, opened 2026-08-17
feat(trading-strategies): add indicator signals

### LINKED ISSUE #1310: feat(trading-strategies): add a declarative indicatorSignal adapter
## Motivation

Strategies currently have to own and orchestrate technical indicators themselves:

1. Create the indicator.
2. Create or select a candle batcher.
3. Feed each completed candle into the indicator.
4. Track warm-up and readiness.
5. Read the latest result before making a decision.

This mixes data plumbing with decision logic. For example, an SMA crossover strategy should describe the SMA readings it needs, rather than manually feeding candles into multiple batchers and indicators.

## Proposal

Add an `indicatorSignal()` adapter that declaratively connects a runtime-managed source to any incremental `trading-signals` indicator:

```ts
const sma = indicatorSignal({
  id: 'sma-20',
  source: bars5m,
  createIndicator: () => new SMA(20),
  selectInput: bar => bar.close.toNumber(),
});
```

The returned value is a cold signal definition. Calling `indicatorSignal()` should not instantiate the indicator or subscribe to the source.

When a decision runtime loads the definition, it should:

- Create one indicator instance per strategy session.
- Subscribe the instance to `bars5m`.
- Call `selectInput()` for every completed bar.
- Pass appended values to `indicator.add()`.
- Pass revised values to `indicator.replace()`.
- Track warm-up through `indicator.isStable`.
- Publish typed, immutable readings to the strategy.

The strategy would only consume the reading:

```ts
const reading = context.signals.get(sma);

if (reading.status !== 'ready') {
  return;
}

console.log(reading.value);
```

## Suggested API

```ts
export interface IncrementalIndicator<Input, Result> {
  readonly isStable: boolean;

  add(input: Input): Result | null;
  replace(input: Input): Result | null;
  getRequiredInputs(): number;
  getResult(): Result | null;
}

export interface IndicatorSignalOptions<SourceValue, Input, Result> {
  /** Stable name for diagnostics, traces, configuration, and optional persistence. */
  id: string;

  /** Runtime-managed source whose values advance the indicator. */
  source: SignalDefinition<SourceValue>;

  /** Mutable indicators must not be shared between sessions or backtests. */
  createIndicator(): IncrementalIndicator<Input, Result>;

  /** Convert the source value into the input expected by the indicator. */
  selectInput(value: SourceValue): Input;
}

export function indicatorSignal<SourceValue, Input, Result>(
  options: IndicatorSignalOptions<SourceValue, Input, Result>,
): SignalDefinition<Result>;
```

The adapter should be generic rather than candle-specific. Candles are the primary use case, but any compatible source could be mapped into an indicator input.

For example, structured candle input should work too:

```ts
const atr = indicatorSignal({
  id: 'atr-14',
  source: bars1h,
  createIndicator: () => new ATR(14),
  selectInput: bar => ({
    close: bar.close.toNumber(),
    high: bar.high.toNumber(),
    low: bar.low.toNumber(),
  }),
});
```

## Reading states

The adapter should distinguish warm-up from missing, stale, or failed data:

```ts
type SignalReading<T> =
  | {
      status: 'warming';
      receivedInputs: number;
      requiredInputs: number;
    }
  | {
      status: 'ready';
      value: T;
      effectiveAt: number;
      revision: number;
    }
  | {
      status: 'stale' | 'unavailable' | 'error';
      reason?: string;
      lastValue?: T;
    };
```

This prevents `null` from representing several different runtime conditions.

## Update semantics

Updates should be transactional:

1. Update all affected source streams.
2. Update every dependent indicator.
3. Commit one immutable signal snapshot.
4. Evaluate the strategy once.

At a five-minute boundary, this ensures the strategy cannot observe a new fast indicator alongside a slow indicator that has not yet processed the same boundary.

Multiple signals consuming the same candle source should share the source and its batching work while retaining independent indicator instances.

## Non-goals

- `indicatorSignal()` does not decide whether a result means buy or sell.
- It does not create an `OrderAdvice`.
- It does not fetch or batch candles itself.
- It does not share mutable indicator instances between sessions.
- It does not require indicator results to be bullish/bearish; raw numeric and structured results remain available.

## Acceptance criteria

- [ ] `indicatorSignal()` can wrap numeric-input indicators such as `SMA`.
- [ ] It can wrap structured-input indicators such as `ATR`.
- [ ] Indicators are instantiated separately for every runtime session.
- [ ] Appended source values call `add()`.
- [ ] Source revisions call `replace()`.
- [ ] Readings remain `warming` until the indicator is stable.
- [ ] The latest result is exposed as a typed, immutable reading.
- [ ] Multiple indicators can consume one shared candle-bar source.
- [ ] Strategies do not need to call `add()`, `replace()`, or `CandleBatcher` themselves.
- [ ] Live trading and backtesting use the same update path.
- [ ] Tests demonstrate migrating `SmaCrossoverStrategy` without changing its advice sequence.

## Package placement

This should initially live in `trading-strategies`, where candle sources and the decision runtime are orchestrated. It can consume `trading-signals` indicators through the small structural `IncrementalIndicator` interface, keeping the core indicator package independent of strategy-runtime and exchange concerns.

sergio12S:
## Summary

- add a cold, generic `indicatorSignal()` definition and per-session signal runtime
- centralize shared candle batching and publish transactional signal snapshots
- migrate `SmaCrossoverStrategy` to consume typed readings without changing its advice sequence

`SignalRuntime` creates independent mutable indicator instances for each
strategy session. It maps appended source values to `add()`, revisions to
`replace()`, and exposes frozen typed readings that distinguish warming, ready,
stale, unavailable, and error states. Candle signals are batched once per
runtime even when multiple indicators consume the same timeframe. All affected
indicators are advanced before a new snapshot is published, so a strategy
cannot observe a partially updated boundary.

The SMA crossover migration keeps its existing advice semantics and uses the
same `updateCandle()` path in live sessions and backtests. Focused tests cover
cold definitions, session isolation, numeric and structured inputs, revisions,
readiness states, and shared-source batching.

Closes #1310.

## Verification

- `npm run lint`
- `npm test` (160 test files, 1740 tests)

### DIFF
--- a/packages/trading-strategies/src/index.ts
+++ b/packages/trading-strategies/src/index.ts
@@ -1,4 +1,5 @@
 export * from './backtest/index.js';
+export * from './signal/index.js';
 export * from './strategy/index.js';
 export * from './trader/index.js';
 export {BuyOnceStrategy, BuyOnceSchema, type BuyOnceConfig} from './strategy-buy-once/BuyOnceStrategy.js';
--- a/packages/trading-strategies/src/signal/CandleSignal.ts
+++ b/packages/trading-strategies/src/signal/CandleSignal.ts
@@ -1,0 +1,23 @@
+import type {BatchedCandle} from '@typedtrader/exchange';
+import type {SignalDefinition} from './SignalDefinition.js';
+
+export interface CandleSignalOptions {
+  id: string;
+  intervalInMillis: number;
+}
+
+export interface CandleSignalDefinition extends SignalDefinition<BatchedCandle> {
+  readonly intervalInMillis: number;
+  readonly kind: 'candle';
+}
+
+export function candleSignal(options: CandleSignalOptions): CandleSignalDefinition {
+  if (!Number.isFinite(options.intervalInMillis) || options.intervalInMillis <= 0) {
+    throw new Error('A candle signal interval must be a positive number of milliseconds');
+  }
+  return Object.freeze({
+    id: options.id,
+    intervalInMillis: options.intervalInMillis,
+    kind: 'candle',
+  });
+}
--- a/packages/trading-strategies/src/signal/SignalDefinition.ts
+++ b/packages/trading-strategies/src/signal/SignalDefinition.ts
@@ -1,0 +1,78 @@
+declare const signalValue: unique symbol;
+
+export interface SignalDefinition<Value> {
+  readonly id: string;
+  readonly kind: 'source' | 'candle' | 'indicator';
+  readonly [signalValue]?: Value;
+}
+
+export interface SourceSignalOptions {
+  id: string;
+}
+
+export interface SourceSignalDefinition<Value> extends SignalDefinition<Value> {
+  readonly kind: 'source';
+}
+
+export function sourceSignal<Value>(options: SourceSignalOptions): SourceSignalDefinition<Value> {
+  return Object.freeze({id: options.id, kind: 'source'});
+}
+
+export interface IncrementalIndicator<Input, Result> {
+  readonly isStable: boolean;
+
+  add(input: Input): Result | null;
+  replace(input: Input): Result | null;
+  getRequiredInputs(): number;
+  getResult(): Result | null;
+}
+
+export interface IndicatorSignalOptions<SourceValue, Input, Result> {
+  createIndicator(): IncrementalIndicator<Input, Result>;
+  id: string;
+  selectInput(value: SourceValue): Input;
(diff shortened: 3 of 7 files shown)


bennycode:
Hey @sergio12S, what problem is this PR actually solving?

bennycode:
@sergio12S, are you still interested in shipping this?

bennycode:
Closed because of inactivity.

PR #1316, opened 2026-08-17
feat(trading-strategies): add performance risk metrics

### LINKED ISSUE #1196: feat(trading-strategies): add Sharpe, Sortino, and max drawdown to PerformanceCalculator
## Problem

`PerformanceCalculator` currently reports ROI, buy-and-hold baseline, win rate, streaks, and trade counts — but no risk-adjusted metrics. Two strategies with the same ROI can carry wildly different risk, and right now backtest results give no way to tell them apart.

## Proposal

Extend the `performance` summary in `BacktestResult` with:

- **Sharpe ratio** — mean return / stddev of returns over the equity curve
- **Sortino ratio** — same, but penalizing downside deviation only
- **Max drawdown** — largest peak-to-trough decline of portfolio value (percent)

This requires tracking an equity curve (portfolio value per candle or per trade) inside `BacktestExecutor`, which `PerformanceCalculator` can then derive all three metrics from.

## Notes

- Cheap to implement, immediately makes backtest results comparable across strategies.
- Surface the new metrics in the docs backtester stat cards (`BacktestResults`) as a follow-up.
- Prerequisite for parameter optimization: an optimizer needs a risk-adjusted fitness function, not raw ROI.

sergio12S:
## Summary

- track portfolio equity at the initial open and after every processed candle
- add Sharpe ratio, Sortino ratio, and maximum drawdown to the backtest performance summary
- include held balances in equity so pending orders do not create artificial drawdowns

The ratios use consecutive candle-to-candle portfolio returns and remain
unannualized because the executor accepts arbitrary candle intervals. Sortino
uses zero as its target return, while maximum drawdown is reported as a positive
peak-to-trough percentage. Degenerate inputs return finite zeroes instead of
`NaN` or infinity.

Tests cover closed-form metric examples, flat and insufficient curves, the
executor's candle-by-candle integration, and valuation while an order balance
is held.

Closes #1196.

## Verification

- `npm run lint`
- `npm test` (160 test files, 1740 tests)

### DIFF
--- a/packages/trading-strategies/src/backtest/BacktestExecutor.ts
+++ b/packages/trading-strategies/src/backtest/BacktestExecutor.ts
@@ -38,6 +38,9 @@ export class BacktestExecutor {
     const trades: BacktestTrade[] = [];
     const skippedAdvices: BacktestSkippedAdvice[] = [];
     let totalFees = new Big(0);
+    const firstOpenPrice = candles.length > 0 ? new Big(candles[0].open) : new Big(0);
+    const initialPortfolioValue = initialBaseBalance.mul(firstOpenPrice).plus(initialCounterBalance);
+    const equityCurve = [initialPortfolioValue];
 
     for (const candle of candles) {
       // Step 1: Match pending orders from previous iteration against this candle
@@ -67,6 +70,7 @@ export class BacktestExecutor {
       const batchedCandle = CandleBatcher.createOneMinuteBatchedCandle([candle]);
       const state = await this.#buildState(tradingPair, tradingRules, feeRates);
       const advice = await strategy.onCandle(batchedCandle, state);
+      equityCurve.push(await this.#calculatePortfolioValue(new Big(candle.close)));
 
       if (!advice) {
         continue;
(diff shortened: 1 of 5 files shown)


packages/trading-strategies/src/backtest/BacktestExecutor.ts
```diff
@@ -99,14 +103,18 @@ export class BacktestExecutor {
...
     const finalBalances = await exchange.getAvailableBalances(tradingPair);
-    const firstOpenPrice = candles.length > 0 ? new Big(candles[0].open) : new Big(0);
     const lastClosePrice = candles.length > 0 ? new Big(candles[candles.length - 1].close) : new Big(0);
 
-    const initialPortfolioValue = initialBaseBalance.mul(firstOpenPrice).plus(initialCounterBalance);
     const finalPortfolioValue = finalBalances.base.mul(lastClosePrice).plus(finalBalances.counter);
     const profitOrLoss = finalPortfolioValue.minus(initialPortfolioValue);
 
-    const performance = this.#buildPerformanceSummary(trades, candles, initialPortfolioValue, finalPortfolioValue);
+    const performance = this.#buildPerformanceSummary(
```
bennycode: todo: let's create a config object here

bennycode:
@sergio12S thank you for your contribution! I found and fixed a bug in the Sharpe/Sortino ratios and committed it to your PR. Will merge it asap. Best, Benny

PR #1317, opened 2026-08-21
feat(exchange): add agent-friendly broker CLI

bennycode:
## What

A broker-generic `exchange-cli` binary covering account inspection and the full order lifecycle on any supported broker, plus four real bugs found by driving every command against live broker APIs while building it.

### CLI

```sh
exchange-cli quote RRl_EQ --broker trading212 --counter GBX
exchange-cli buy RRl_EQ 1 --broker trading212 --limit 1000 --dry-run
exchange-cli buy RRl_EQ 1 --broker trading212 --limit 1000
exchange-cli wait RRl_EQ <orderId> --broker trading212 --counter GBX --timeout 5m
exchange-cli watch-candles AAPL --broker alpaca --counter USD --take 3 --idle 5m
```

- Commands: `verify`, `balances`, `instruments`, `quote`, `rules`, `orders`, `fills`, `buy`/`sell` (market or `--limit`, with `--dry-run` validating against trading rules and estimating fees without placing), `wait` (poll an order to its terminal state), `cancel`, `candles`, `watch-candles`/`watch-orders` (NDJSON streaming, `--take` to exit after n events), `time`.
- Built for coding agents as much as humans: JSON on stdout, broker problem types verbatim on stderr, exit code 1 on failure, paper by default with `--live` opt-in.
- Runs through the real composition root (`getBrokerClient`), so every invocation exercises the production path. Credentials via `<BROKER>_API_KEY`/`_API_SECRET` env vars, loaded from `.env.live` when `--live` is passed and `.env.sandbox` otherwise — the file is the environment, so a machine without `.env.live` cannot trade the real account. Trading212 candle commands source market data from Alpaca (`ALPACA_API_KEY` from the same file; always the production data hosts, since the sandbox data hosts are Broker-API-partner-only and 401 regular keys).
- Streaming guard rails: Alpaca serves one market-data WebSocket per API key and a second connection makes the server drop the first, so a concurrent stream on the same machine now fails fast naming the holding PID (per-key lock file, key id hashed, never the secret), and `--idle` turns a subscribed-but-silent stream into a quick diagnostic failure instead of an indefinite hang. The bin exits explicitly after flushing output because the broker WebSockets deliberately stay open for long-running sessions.
- Replaces the `trading212:buy` / `trading212:listBalances` demo scripts.

### Fixes (each found live)

- **Trading212 market orders on non-US venues**: the broker hardcoded `extendedHours: true`, which LSE instruments reject with 400. Market orders now retry once on regular hours when the venue rejects the flag.
- **Trading212 limit orders were broken everywhere**: the limit endpoint rejects any request carrying `extendedHours` with a generic 400 "Invalid payload" (verified on demo and live, US and LSE). The field is removed from the request schema.
- **`estimateFee` misstated cross-currency fees ~100x**: it multiplied a counter-currency notional by the fee rate and labeled the result with the account currency, turning a ~10 GBP trade's FX fee of 1.5 GBX into "EUR 1.50". Estimates are now denominated in `pair.counter`; the realised debit currency still comes back on `Fill.feeAsset`. The now unused `getFeeAsset` hook is removed (no consumers).
- **Alpaca live cash accounts failed schema validation**: `daytrade_count` and `pattern_day_trader` are omitted on some live accounts and are now `nullish` (nothing consumes them).

## Verification

142 tests (37 new), typecheck and lint clean. Every command was additionally exercised against the live APIs: Trading212 paper full order lifecycle including `wait` timeout and cancel, Alpaca live quote/candles/streaming, lock contention between two concurrent streams (loser refused instantly naming the holder PID, winner streamed undisturbed), and `--idle` firing after 5s of silence.


bennycode:
Addressed the review in 78fc270:

- Stale lock takeover is now atomic: the dead holder's file is removed and the wx create is re-raced, so two contenders that both saw a stale lock can't proceed together. Unreadable lock content counts as stale.
- --idle is validated before the lock is acquired, so a parse error can't leak the lock file (test pins that the lock is never touched on invalid input).
- Duration parse errors name the failing flag (--idle, --timeout, --poll) instead of always blaming --interval.
- The closing watch-* summary prints compact, keeping stdout one JSON object per line end to end.
- Usage text spells out both env var names (TRADING212_PAPER_API_KEY and TRADING212_PAPER_API_SECRET) instead of the ambiguous shorthand.

The CodeQL finding is dismissed as a false positive: the SHA-256 fingerprints the API key id (not the secret) purely to name a lock file without leaking the credential. It's never stored or compared as an authentication verifier, so password-KDF guidance doesn't apply.

PR #1319, opened 2026-08-22
feat(trading-signals): runtime input shapes and a generic indicator CLI

bennycode:
## What

Two connected changes to the `trading-signals` package:

### 1. Every indicator declares its input shape at runtime

TypeScript input types are erased at runtime, so a generic consumer (a CLI, the visual strategy builder, docs tooling) cannot know whether an indicator consumes a close series or full candles — and feeding the wrong shape silently produces NaN rather than an error.

Each indicator now carries `inputShape`, a runtime value from the new `IndicatorInputShape` constant (`value`, `high-low`, `high-low-close`, `high-low-close-volume`, `open-high-low-close`, `open-high-low-close-volume`). The declaration is **compiler-verified**: `InputShapeOf<Input>` maps each class's `Input` generic to the one correct literal, so declaring a wrong shape (or forgetting one on a new indicator) fails to compile. `MovingAverage` subclasses inherit `value` from the family base; 117 classes declare explicitly.

### 2. A generic `trading-signals` CLI

```sh
trading-signals atr 14 --csv candles.csv
cat candles.json | trading-signals rsi 14
trading-signals stochasticoscillator kPeriod=5 dPeriod=3 kSlowingPeriod=3 --csv candles.csv
```

- The indicator is resolved generically from the package exports (lowercased class name), so the CLI never imports individual indicators — new ones are available the moment they're exported. Abstract bases are filtered out by checking that the prototype chain actually carries an `update` implementation.
- Input formats auto-detected: CSV with header, NDJSON, or a JSON array; stdin or `--csv`. Field aliases `o/h/l/c/v` accepted; `time`/`date`/`timestamp` passed through.
- The declared `inputShape` decides which candle fields are fed — closes for value-series indicators, the exact candle fields for the rest, with a clear error naming any missing field.
- Constructor parameters from the command line: numbers positionally, `key=value` pairs collected into one config object, covering both constructor styles in the library.
- Output streams one JSON line per stable result as candles arrive, so the CLI composes with live candle feeds over a pipe; `--last` prints only the final value.
- Zero new dependencies (`node:util` + `node:readline`); the bin is excluded from coverage since it only wires stdin/stdout to the tested `runCli` core.

## Verification

1760 tests (21 new for the CLI, with exact pinned values per input shape), 100% coverage thresholds passing, typecheck and lint clean. The built bin verified against CSV files, piped NDJSON, and piped JSON arrays.


bennycode:
Addressed the review in 6a93453:

- New VOLUME input shape: VROC, PVO, and RVOL declare it, and the CLI feeds the volume field instead of the close. The scalar branch of InputShapeOf allows VALUE or VOLUME since the type system cannot tell a price series from a volume series — that distinction is the one part a human declares. Regression: constant closes with quadrupling volume must register as +300% VROC.
- MACD and BollingerBandsWidth construct through factories that wire their indicator dependencies from flat numeric parameters (macd 12 26 9); missing parameters fail with a usage message instead of crashing at first update.
- Dotted config keys nest, so pgo interval=14 signalThresholds.overbought=4 reaches the nested config instead of being silently dropped.
- CSV parsing is quote-aware (commas inside quotes, "" escapes) with a regression test.
- Empty CSV cells are rejected instead of passing as 0 via Number('').
- The bin accepts both --csv path and --csv=path.
- Assertion rationales moved from comments into expect messages per the repo testing rule.

1766 tests, 100% coverage thresholds passing, lint and typecheck clean.

bennycode:
Superseded by #1364, which carries the first half of this forward.

The input shapes were the valuable part and they are in #1364, re-targeted at current main: every indicator declares one, `InputShapeOf<Input>` derives it from the input type, and the CLI reads the declaration instead of inferring it. Two changes since this branch: `value` became `PRICE` alongside a restored `VOLUME`, so a plain number says which series it is, and a test holds each declaration against the name of the parameter the indicator reads.

The second half, the CLI inside this package, is dropped. A command line now ships as `trading-signals-cli` (#1361), a package of its own, which is what keeps `trading-signals` free of dependencies and of a `bin`.

Closing in favour of #1364.

PR #1351, opened 2026-09-02
feat(@typedtrader/exchange): estimate Alpaca regulatory fees on equity sells

bennycode:
Stacked on #1350 — review that one first, this targets its branch.

Alpaca passes through the SEC Section 31 fee and the FINRA Trading Activity Fee on equity sells. Both are now estimated per fill, so stock fills stop reporting a flat zero.

## The fee is not rounded per trade

Alpaca sums a whole day per fee type and rounds *that* up to the penny once. On a live account, 48 shares across nine sells on 2024-07-19 were billed $0.01 of TAF in total, where per-trade rounding reports $0.09. Leaving each trade unrounded keeps a day's estimates summing to within one penny of the real bill, which is the best attribution available — equity `FEE` activities carry no order, symbol or quantity.

## The SEC rate is not a constant

It is re-set annually to collect exactly the Commission's appropriation, so it moves a lot and sat at $0.00 from 2025-05-14 to 2026-04-04. A hardcoded rate would have invented fees for sixteen months. `SEC_FEE_RATES` is a dated table covering FY2020 through FY2026, each window read out of the SEC's own Section 31 order, and it reconciles against every `REG` and `TAF` activity on the live account it was derived from. The zero window shows up there as sell days carrying a TAF charge and no REG charge at all.

The table gains one row a year. Adding it by hand from the fee rate advisories when a new fiscal year takes effect is less machinery than fetching and parsing the orders at runtime, and it keeps a fill priced the same on every run.

## Crypto fees round up too

A live crypto BUY settled a question this module had documented as unknowable. A market BUY of 0.001254903 BTC at the 0.25% taker rate computes to 0.0000031372575 BTC; Alpaca billed 0.000003138 and the position grew by exactly the difference. Base-denominated fees round up in their own unit, the same way fiat fees round up to the penny, so the rule is uniform.

That trade also showed Alpaca bills **per execution, not per order**: it filled in two legs and drew two separately rounded `CFEE` entries. Computing once over a whole order can come out up to one increment per leg low. Both legs are pinned as tests.

## Not included

CAT applies to both sides and is billed the same day-aggregated way, but its per-share rate is too small to pin down from observed activity, so it is left out rather than guessed at.


packages/exchange/src/broker/alpaca/AlpacaFees.ts
```diff
@@ -3,17 +3,64 @@ import {OrderSide, type FeeRate, type OrderType} from '../Broker.js';
...
+  {from: '2023-02-27', rate: new Big('0.0000080')},
+  {from: '2024-05-22', rate: new Big('0.0000278')},
+  {from: '2025-05-14', rate: new Big('0')},
+  {from: '2026-04-04', rate: new Big('0.0000206')},
+];
+
+/**
+ * FINRA Trading Activity Fee, per share sold and capped per trade. Dated for the day FINRA moves it.
+ *
+ * @see https://www.finra.org/finra-data/browse-catalog/trading-activity-fee
```
bennycode: website doesn't exist

bennycode: Fixed — the catalog path 404s, `finra.org/rules-guidance/guidance/trading-activity-fee` is the live page. I swept every other link in the diff; the rest resolve, and the sec.gov one returns 403 to automated requests but the page is correct.

PR #1352, opened 2026-09-04
fix(@typedtrader/exchange): parse Alpaca cash accounts

bennycode:
`AccountSchema` required `daytrade_count` and `pattern_day_trader`, but Alpaca omits both on cash accounts (`multiplier: "1"`) rather than reporting zero — Pattern Day Trader tracking only applies to margin trading.

So `getAccount()` threw a `ZodError` against a live cash account:

```
ZodError: [{ "expected": "number", "code": "invalid_type",
             "path": ["daytrade_count"],
             "message": "Invalid input: expected number, received undefined" }]
```

That also took down `listBalances()`, which is `getAccount()`'s only caller.

## Why optional is the right fix rather than a default

Neither field is read anywhere in the codebase — they exist only in the schema. Two unused required fields were failing the whole call. Making them optional matches what the API actually guarantees; giving them a default would fabricate a day-trade count that Alpaca never sent.

The rest of the schema is untouched, and a test asserts that a response missing a field every account *does* report still fails to parse, so this doesn't turn into a blanket loosening.

## Verification

Against a live cash account, `getAccount()` and `listBalances()` both succeed where they previously threw:

```
getAccount() OK: cash=$225.74 currency=USD multiplier=1
  daytrade_count=undefined  pattern_day_trader=undefined
listBalances() OK: 7 balances
```

Found while working on an unrelated change; it is independent of #1350 and #1351 and branches from `main`.

### DIFF
--- a/packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
+++ b/packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
@@ -1,12 +1,7 @@
 import {describe, expect, it} from 'vitest';
-import {AccountSchema} from './AccountSchema.js';
+import {AccountSchema, type Account} from './AccountSchema.js';
 
-/**
- * Shape of a live cash account, trimmed to the fields the schema declares. Alpaca returns no
- * `daytrade_count` and no `pattern_day_trader` here: both belong to Pattern Day Trader tracking,
- * which only applies to margin accounts.
- */
-const CASH_ACCOUNT = {
+const CASH_ACCOUNT: Account = {
   account_blocked: false,
   account_number: '245695408',
   buying_power: '325.7',
@@ -27,7 +22,7 @@ const CASH_ACCOUNT = {
   trade_suspended_by_user: false,
   trading_blocked: false,
   transfers_blocked: false,
-} as const;
+};
 
 describe('AccountSchema', () => {
   it('parses a cash account that omits the Pattern Day Trader fields', () => {
--- a/packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
+++ b/packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
@@ -1,0 +1,61 @@
+import {describe, expect, it} from 'vitest';
+import {AccountSchema} from './AccountSchema.js';
+
+/**
+ * Shape of a live cash account, trimmed to the fields the schema declares. Alpaca returns no
+ * `daytrade_count` and no `pattern_day_trader` here: both belong to Pattern Day Trader tracking,
+ * which only applies to margin accounts.
+ */
+const CASH_ACCOUNT = {
+  account_blocked: false,
+  account_number: '245695408',
+  buying_power: '325.7',
+  cash: '325.7',
+  created_at: '2023-08-08T18:58:27.267Z',
+  currency: 'USD',
+  equity: '425.7',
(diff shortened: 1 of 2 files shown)


packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
```diff
@@ -0,0 +1,61 @@
+import {describe, expect, it} from 'vitest';
+import {AccountSchema} from './AccountSchema.js';
+
+/**
+ * Shape of a live cash account, trimmed to the fields the schema declares. Alpaca returns no
+ * `daytrade_count` and no `pattern_day_trader` here: both belong to Pattern Day Trader tracking,
+ * which only applies to margin accounts.
+ */
+const CASH_ACCOUNT = {
```
bennycode: add type

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
```diff
@@ -0,0 +1,61 @@
+import {describe, expect, it} from 'vitest';
+import {AccountSchema} from './AccountSchema.js';
+
+/**
```
bennycode: remove comment

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.ts
```diff
@@ -10,15 +10,20 @@ export const AccountSchema = z.looseObject({
   cash: z.string(),
   created_at: z.string(),
   currency: z.string(),
-  daytrade_count: z.number(),
+  /**
```
bennycode: remove comment

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.ts
```diff
@@ -10,15 +10,20 @@ export const AccountSchema = z.looseObject({
...
+  daytrade_count: z.number().optional(),
   equity: z.string(),
   id: z.string(),
   initial_margin: z.string(),
   last_equity: z.string(),
   long_market_value: z.string(),
   maintenance_margin: z.string(),
   multiplier: z.string(),
-  pattern_day_trader: z.boolean(),
+  /** Omitted on cash accounts for the same reason as {@link AccountSchema} `daytrade_count`. */
```
bennycode: remove comment

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
bennycode: Typed as `Account`, so the fixture is checked against the schema instead of inferred.

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.test.ts
bennycode: Removed.

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.ts
bennycode: Removed. The rationale stays in the commit that made the field optional.

packages/exchange/src/broker/alpaca/api/schema/AccountSchema.ts
bennycode: Removed.

